PR #1, opened 2016-06-12
Upgrade remoting to 2.62

oleg-nenashev:
This fix upgrades remoting to 2.60, which is the latest version.

There are much changes since 2.52. The major ones are:
* Support of the JNLP3 protocol with encryption
* Memory leak fixes
* Fixes for several connectivity issues

Incomplete changelog is available here: https://github.com/jenkinsci/remoting/blob/stable-2.x/CHANGELOG.md

@reviewbybees @jenkinsci/code-reviewers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@
 
 FROM java:8-jdk
 
-RUN curl --create-dirs -sSLo /usr/share/jenkins/slave.jar http://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/2.60/remoting-2.60.jar \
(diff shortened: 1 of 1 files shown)


carlossg:
I think this repo needs to be removed, in favor of https://github.com/jenkinsci/docker-jnlp-slave @ndeloof

ndeloof:
docker-jnlp-slave should be built on top of this one and only add jnlp specific stuff

oleg-nenashev:
Well, here we cannot disable JNLP3 protocol

PR #4, opened 2017-03-11
add Group flag

joppa27:
just added the missing `-G` to the adduser command.
Good catch by  @jwieringa

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ MAINTAINER Nicolas De Loof <nicolas.deloof@gmail.com>
 
 ENV HOME /home/jenkins
 RUN addgroup -S -g 10000 jenkins
-RUN adduser -S -u 10000 -h $HOME jenkins jenkins
+RUN adduser -S -u 10000 -h $HOME -G jenkins jenkins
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
@carlossg @dduportal any concerns regarding the downstream images?

dduportal:
@oleg-nenashev @joppa27 Good catch ! 

Improvement proposal: following the method used on the officiel jenkins image (https://github.com/jenkinsci/docker/blob/master/Dockerfile#L17 ) , by combining `ARG` and `ENV` instruction to providable customizable agent and group, thus avoiding this kind of confusion in the future.

It also let the end user to use the mapping they want ( the label is just a facility: uid and gid are much needed since the jenkins docker usages tend to bind mount volume to the host).

dduportal:
@oleg-nenashev but you can merge this here, got my 🐝  (and my axe :p )

oleg-nenashev:
Let's see how the automatic build flies

dduportal:
Thanks for the contribution @joppa27 and @jwieringa ! This is much welcome !

PR #6, opened 2017-03-15
Create README for the repository

oleg-nenashev:
This PR creates a Readme for the current state of the DockerHub. `2.62` branches need to be revived in the repository since we have automatic builds now. Maybe it's better to rename them to something like `2.x` and `2.x-alpine` since @jglick wants to change the version naming approach in Remoting 2.x

screen shot 2017-03-15 at 10 17 34

@reviewbybees, esp. @ndeloof and @carlossg

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,22 @@
+Jenkins Agent Docker image
+===
+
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
@reviewbybees done

PR #12, opened 2017-06-26
Create Jenkins workspace folder

sboschman:
Only way I have been able to mount a Docker volume (docker volume create jenkins-workspace) from the host as workspace directory with the right permissions. Without an existing workspace directory owned by the Jenkins user Docker mounts the workspace directory as root and hence the Jenkins users lacks permissions to use the workspace directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,6 @@ RUN apk add --update --no-cache curl bash git openssh-client openssl \
   && apk del curl
 
 USER jenkins
-RUN mkdir /home/jenkins/.jenkins
+RUN mkdir /home/jenkins/.jenkins /home/jenkins/workspace
 VOLUME /home/jenkins/.jenkins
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
I think this change should be delivered along with enabling Remoting workspaces by default in Docker agents, see [JENKINS-39370](https://issues.jenkins-ci.org/browse/JENKINS-39370) and [JENKINS-44114](https://issues.jenkins-ci.org/browse/JENKINS-44114). It adds way more justification to it though likely it has to be a new docker since logs on the disk are not always required in Docker.

oleg-nenashev:
@sboschman IMHO it is superseded by #13, WDYT?
The only difference is that directory creates `/home/jenkins/agent`, so not sure it 100% fits your use-case

oleg-nenashev:
Closing in favor #13. Please reopen if it is not enough

Issue #7, opened 2017-03-15
Make group/user customizable via ARG/ENV


oleg-nenashev:
Just to note the proposal by @dduportal in https://github.com/jenkinsci/docker-slave/pull/4#issuecomment-286697425

slide:
Is this resolved now?

timja:
appears so

PR #25, opened 2018-04-06
Fixing #7 - Make group/user customizable via ARG/ENV

egorse:
The args taking from https://github.com/jenkinsci/docker but default values are kept original (i.e. 10000).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,22 +23,27 @@
 FROM openjdk:8-jdk
 MAINTAINER Oleg Nenashev <o.v.nenashev@gmail.com>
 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
@egorse Do you need releases of all branches/tags + https://github.com/jenkinsci/docker-jnlp-slave ?

egorse:
@oleg-nenashev tags would be nice - jnlp-slave is up to you

soujiro32167:
This would be very useful for me, since my slaves are running into permissions issues when executing docker commands. Same as this issue: https://github.com/jenkinsci/docker/issues/263

PR #34, opened 2018-11-06
[JENKINS-51986] - Introduce a Java 11 packaging for the Debian image

oleg-nenashev:
It creates a branch for Java 11 and Remoting. Apparently works as is

https://issues.jenkins-ci.org/browse/JENKINS-51986


oleg-nenashev:
@carlossg `latest-jdk11` according to #35

PR #36, opened 2018-11-09
Avoid including an image tag in the example as it will just be outdated

jglick:


oleg-nenashev:
Thanks @jglick abd @carlossg !

PR #38, opened 2018-12-01
Fixed Jenkins UID and GID to be consistent with the Jenkins Docker image

### LINKED ISSUE #32: uid/gid : 10 000 or 1 000
ARG uid=10000
ARG gid=10000

Not 1000, like the uid jenkins in the docker image jenkins/jenkins ?

robvadai:
The Jenkins UID and GID has to be same as the Jenkins Docker image otherwise Docker-in-Docker can not be used, by mounting `/var/run/docker.sock`.

Jenkins Docker image settings: https://github.com/jenkinsci/docker/blob/master/Dockerfile#L7

Resolves #32

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,8 +25,8 @@ MAINTAINER Oleg Nenashev <o.v.nenashev@gmail.com>
 
 ARG user=jenkins
 ARG group=jenkins
-ARG uid=10000
-ARG gid=10000
+ARG uid=1000
+ARG gid=1000
 
 ENV HOME /home/${user}
 RUN groupadd -g ${gid} ${group}


batmat:
Should it be fixed too in other branches?

BTW, I'm starting to think all this should be refactored into a single branch so PRs can be done to address all flavors together. The very same issue just surfaced on #39 too.

carlossg:
yes, we should remove the branches, they are legacy from the Docker official images times

robvadai:
Hi guys, is this PR going to be merged?

batmat:
@robvadai I don't think so.  Sorry I was unclear in my last comment: I think we should keep all variants aligned, so please file the equivalent PR against the other branches too so that it can be considered.

Thank you!
PS: started to think recently that actually the right fix would be to make the id configurable when running the container.

robvadai:
The thing is we should be pragmatic, so your suggestions are good but I can't see merging branches and making the UID configurable happening any time soon. Which means the only option to get this image working is to use a fork, like I did.

Have you got an ETA for for merging branches and make the UID configurable in runtime?

batmat:
(I guess I shouldn't have put that _PS_, that made things unclear. Was more a loud thinking also for other maintainers or anyone interested to contribute.)

For the context in this very PR, I'm saying I think there should two others so that it's mergeable, so that we keep all agent's images UIDs consistent.
You filed the PR against `master`, there should be one too against the `alpine` and the `jdk11` branches.

image

I hope it's clearer.

robvadai:
OK no worries.

Alpine branch: https://github.com/jenkinsci/docker-slave/pull/42
JDK11 branch: https://github.com/jenkinsci/docker-slave/pull/43

oleg-nenashev:
Usually porting of the PR is being done by the maintainers. But thanks for the PRs

robvadai:
@oleg-nenashev thank you for merging!
And a Merry Christmas in advance 🎄

PR #39, opened 2018-12-03
Install Git LFS in image.

kleini:
See https://github.com/jenkinsci/docker-jnlp-slave/issues/70

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ ARG VERSION=3.27
 
 ARG AGENT_WORKDIR=/home/${user}/agent
 
-RUN apk add --update --no-cache curl bash git openssh-client openssl procps \
+RUN apk add --update --no-cache curl bash git git-lfs openssh-client openssl procps \
   && curl --create-dirs -sSLo /usr/share/jenkins/slave.jar https://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/${VERSION}/remoting-${VERSION}.jar \
   && chmod 755 /usr/share/jenkins \
   && chmod 644 /usr/share/jenkins/slave.jar \


oleg-nenashev:
@jenkinsci/java11-support sorry, just testing the review requests by team

TristanFAURE:
Hello 
do you have updates of this PR ?

batmat:
AFAIK no @TristanFAURE, I still think we should keep various flavors aligned.

Given in the meantime I filed https://github.com/jenkinsci/docker-slave/pull/44, we should merge it and then reconsider here I think.

njtman:
Any updates on this one ?
To remain consistent, I think `git-lfs` should also be added to the Dockerfile for Jenkins master.

When Jenkins performs builds of pull request branches, Jenkins master first performs a clone of the repository on Jenkins master. More info [here](https://issues.jenkins-ci.org/browse/JENKINS-33273?focusedCommentId=292273#comment-292273). This is because lightweight checkout is not yet support for PR branches, as discussed [here](https://support.cloudbees.com/hc/en-us/articles/115002991272-Why-is-my-multibranch-project-cloning-the-whole-repository-on-the-master-).  Unfortunately PR branches branches fail to build when `Git LFS pull after checkout` is enabled for a `GitHub Organization` in Jenkins.  Builds fail because a checkout is first performed on Jenkins master and it tries to perform a `git-lfs pull`, however `git-lfs` is not installed causing the build to fail.

Therefore, enabling `Git LFS pull after checkout` for a given `Github Organization` in Jenkins requires that the Jenkins master has git-lfs installed to avoid failing PR builds.

I suppose I have two solutions
1. Maintain my own custom Jenkins master image which has `git-lfs` installed
2. As discussed [here](https://support.cloudbees.com/hc/en-us/articles/115002991272-Why-is-my-multibranch-project-cloning-the-whole-repository-on-the-master-), I could set up a reference repository but that is currently too complicated for my use case.

Instead of having to maintain my own custom Jenkins master image with `git-lfs` installed, I think it would be easier to just install it in the Jenkins master Dockerfile.

What are your thoughts @batmat?

fezno:
@batmat is still anything missing? I was hoping that pull request #52 unblocks this one...

xkobal:
Any update on that merge ?

carlossg:
see #58

PR #40, opened 2018-12-13
Update jenkins remoting to 3.28

emailbob:
https://github.com/jenkinsci/remoting/blob/master/CHANGELOG.md#328


emailbob:
The LABEL part should now be fixed going forward

PR #44, opened 2018-12-22
Merge branches

batmat:
Following up on https://github.com/jenkinsci/docker-slave/pull/38#issuecomment-445197261 and agreement there from @carlossg for instance. 

I propose in this PR to merge all branches into a single one and many Dockerfiles. This will make aligning variants much simpler, instead of having to file 3 PRs like https://github.com/jenkinsci/docker-slave/pull/38.

Also created a small `build.sh` which I'll call from a `Jenkinsfile` as soon as https://issues.jenkins-ci.org/browse/INFRA-1857 gets looked at by @olblak or someone else with sufficient permissions.

**Once this is merged, we will delete the branches and keep only `master` to avoid ambiguity for contributors.**

### DIFF
--- a/Dockerfile-alpine
+++ b/Dockerfile-alpine
@@ -23,6 +23,7 @@
 FROM openjdk:8-jdk-alpine
 MAINTAINER Oleg Nenashev <o.v.nenashev@gmail.com>
 
+ARG VERSION=3.28
 ARG user=jenkins
 ARG group=jenkins
 ARG uid=1000
@@ -31,9 +32,7 @@ ARG gid=1000
 ENV HOME /home/${user}
 RUN addgroup -g ${gid} ${group}
 RUN adduser -h $HOME -u ${uid} -G ${group} -D ${user}
-LABEL Description="This is a base image, which provides the Jenkins agent executable (slave.jar)" Vendor="Jenkins project" Version="3.27"
-
-ARG VERSION=3.27
+LABEL Description="This is a base image, which provides the Jenkins agent executable (slave.jar)" Vendor="Jenkins project" Version="${VERSION}"
 
 ARG AGENT_WORKDIR=/home/${user}/agent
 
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
@batmat Going forward, I believe we should move the build/packaging logic of jenkins/slave and jenkins/jnlp-slave to Remoting. Like https://github.com/jenkinsci/remoting/pull/292 , but something with a new tooling.

There is no sense to keep Docker packaging separated from the main codebase if we switch to the Jenkins-driven builds

batmat:
I agree, and in case you wonder I do remember we already discussed that a
bit some time ago :).

Nevertheless I believe this PR isn't backwards wrt. this mid term
direction: having all variants alongside each other is anyway I think a
good thing to have.

Overall, I thoroughly agree that we should have less agents' repo so it's
easier to manage the same way (building and publishing).

Cheers!


Le dim. 23 déc. 2018 à 01:35, Oleg Nenashev <notifications@github.com> a
écrit :

oleg-nenashev:
Yes, I am OK with the approach in general.
But how do you plan to build version tags? The script does not support it so far. IIUC it needs to be somehow integrated with tags support in Multi-branch Pipeline

Tianny:
Hello, @oleg-nenashev  when to merge the PR [Install Git LFS in image. #39](https://github.com/jenkinsci/docker-slave/pull/39)?  It is almost one month and a half since @batmat referenced the PR on 22 Jan. We need it!

oleg-nenashev:
@Tianny I am not sure what you are asking me about. @batmat requested changes, and they have not been addressed yet. I also do not maintain this repo actively nowadays, please see https://groups.google.com/d/msg/jenkinsci-dev/uc6NsMoCFQI/AIO4WG1UCwAJ

carlossg:
LGTM I'm going to merge before it becomes stale again, and will update dockerhub builds

PR #47, opened 2019-02-09
Add tini to Dockerfile

marco2704:
A few months ago we saw some defunct sh and sleep processes on some of our hosts with Docker 12 where some agents were running on. The solution as described in https://github.com/jenkinsci/docker-jnlp-slave/issues/51 and as you know was using `tini` in the entrypoint.

Since the `--init` option is available in Docker 1.13 or greater, adding `tini` to the agent base image for using it in the https://github.com/jenkinsci/docker-jnlp-slave image may provide more compatibility with older Docker versions as https://github.com/jenkinsci/docker does.


carlossg:
what for?

marco2704:
Same reason as `jenkinsci/docker` uses it. But in this case we would like to use it in `jenkinsci/docker-jnlp-slave`. It mostly comes from the same issue as described at https://github.com/jenkinsci/docker-jnlp-slave/issues/51.

batmat:
I agree with Carlos, this seems to be missing some justification/explanation or a part to make tini actually used?

Can you maybe also enrich the README to explain the usage or so (I suppose it's meant to be used like `tini java -jar ...`?). I understand the goal to correctly cleanup child processes, but I'd prefer to be sure I understand how this is intended to be used.

Thanks

marco2704:
Sorry guys for the missing explanation. A few months ago we saw some defunct sh and sleep processes on some of our hosts with Docker 12 where some agents were running on. The solution as described in https://github.com/jenkinsci/docker-jnlp-slave/issues/51 and as you know was using `tini` in the entrypoint.

Since the `--init` option is available in Docker 1.13 or greater, adding `tini` to the agent base image for using it in the https://github.com/jenkinsci/docker-jnlp-slave image may provide more compatibility with older Docker versions as https://github.com/jenkinsci/docker does.

jglick:
I see no reason for `tini` to be added to this repository without actually being used. It would make more sense in `docker-jnlp-slave`, if actually being used in the `ENTRYPOINT`.

At any rate, while reaping children is in general a good practice, the particular problem reported was I think solved a while back by plugin updates: https://github.com/jenkinsci/durable-task-plugin/pull/49/commits/bd48603aaee0b3cf990d16a594284b5ce99f0f7f

carlossg:
IIRC the problem with defunct processes which required tini in the master image was for long lived containers. 
When the agent container dies all its children will be reaped, so it is only a problem with long lived agents, is that your case?

slide:
@marco2704  Are you still interested in this PR? It has some conflicts that need to be resolved. As was explained above, just adding tini to the docker image doesn't do anything, you need to have tini run the normal entry point so that it is PID 1. Since there is no entry point defined for this container, it really should be in the images that use this as a base that tini would be included.

marco2704:
It seems to me a cross images change to be made. That's why I decided to added to the base image instead of to the "child" images. Let me know whether this would make sense, if not. I will close the PR.

slide:
I think it may be better served in the "child" images and actually implement using tini in the image. That way it's easier to see what it is used for. There are really only two "child" images in the Jenkins world right now, docker-inbound-agent and docker-ssh-agent.

marco2704:
Opening a PR for this in change in docker-inbound-agent and docker-ssh-agent.

PR #48, opened 2019-02-09
Add tini to Dockerfile jdk11

marco2704:
A few months ago we saw some defunct sh and sleep processes on some of our hosts with Docker 12 where some agents were running on. The solution as described in https://github.com/jenkinsci/docker-jnlp-slave/issues/51 and as you know was using `tini` in the entrypoint.

Since the `--init` option is available in Docker 1.13 or greater, adding `tini` to the agent base image for using it in the https://github.com/jenkinsci/docker-jnlp-slave image may provide more compatibility with older Docker versions as https://github.com/jenkinsci/docker does.


oleg-nenashev:
CC @jenkinsci/java11-support

alecharp:
I don't know this plugin enough, but maybe we can provide the `--init` when starting the container, as @MRamonLeon said.

batmat:
FWIW, I commented in https://github.com/jenkinsci/docker-slave/pull/47, which is a similar PR for the `master` branch.

marco2704:
Sorry guys for the missing explanation. A few months ago we saw some defunct sh and sleep processes on some of our hosts with Docker 12. The solution as described in https://github.com/jenkinsci/docker-jnlp-slave/issues/51 and as you know was using `tini` in the entrypoint.

Since the `--init` option is available in Docker 1.13 or greater, adding `tini` to the agent base image for using it in the https://github.com/jenkinsci/docker-jnlp-slave image may provide more compatibility with older Docker versions as https://github.com/jenkinsci/docker does.

carlossg:
this became outdated now that all Dockerfiles are in master

moving discussion to #47

PR #49, opened 2019-02-09
Add tini to Dockerfile alpine

marco2704:
Install `tini` as `jenkinsci/docker` images do in order to child images use it.


carlossg:
I don't think we need tini in the agents

this became outdated now that all Dockerfiles are in master

PR #52, opened 2019-03-15
Install git-lfs to this image, too to match pull request #39

fezno:
Comply to the request of @batmat in pull request #39 to add git-lfs for jdk11 too.


carlossg:
see #58

PR #57, opened 2019-04-29
Upgrade to remoting 3.29

### LINKED ISSUE #55: Slave remote version doesn't match master
Slave remote version is 3.27 while the master version is 3.29, this is causing Jenkins to put those agents on offline and not use them again.

This is urgent please create a slave image for 3.29.

Thank you
image

carlossg:
Fixes #55


carlossg:
on top of #56

PR #61, opened 2019-05-21
Add Stretch Backports source to fix Git LFS install

### LINKED ISSUE #60: Build fails because Git LFS for Debian Stretch requires Backports source
The `openjdk:8-jdk` based image does not build because the `git-lfs` package isn’t found in the default sources. It’s in Backports, so this needs to be added:

```
RUN echo "deb http://deb.debian.org/debian stretch-backports main" >> /etc/apt/sources.list
```

tjanson:
fixes #60

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ LABEL Description="This is a base image, which provides the Jenkins agent execut
 ARG AGENT_WORKDIR=/home/${user}/agent
 
 RUN echo 'deb http://deb.debian.org/debian stretch-backports main' > /etc/apt/sources.list.d/stretch-backports.list
-RUN apt-get update && apt-get install git-lfs
+ RUN apt-get update && apt-get install -t stretch-backports git-lfs
 RUN curl --create-dirs -fsSLo /usr/share/jenkins/slave.jar https://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/${VERSION}/remoting-${VERSION}.jar \
   && chmod 755 /usr/share/jenkins \
   && chmod 644 /usr/share/jenkins/slave.jar
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ LABEL Description="This is a base image, which provides the Jenkins agent execut
 
 ARG AGENT_WORKDIR=/home/${user}/agent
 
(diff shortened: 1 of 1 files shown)


batmat:
retriggering on top of #62 just merged

danielfbm:
could you guys merge this PR? thanks

Dockerfile
```diff
@@ -36,6 +36,7 @@ LABEL Description="This is a base image, which provides the Jenkins agent execut
 
 ARG AGENT_WORKDIR=/home/${user}/agent
 
+RUN echo 'deb http://deb.debian.org/debian stretch-backports main' > /etc/apt/sources.list.d/stretch-backports.list
```
MarkEWaite: I tested this change and it works.  The build that was previously broken by the `apt-get install git-lfs` command now works with this change.

Dockerfile
```diff
@@ -36,6 +36,7 @@ LABEL Description="This is a base image, which provides the Jenkins agent execut
...
 RUN apt-get update && apt-get install git-lfs
```
MarkEWaite: I am surprised that this works, since the [Debian documentation](https://backports.debian.org/Instructions/#index3h2) says that:

> All backports are deactivated by default (i.e. the packages are pinned to 100 by using ButAutomaticUpgrades: yes in the Release files. If you want to install something from backports run:
> 
> `apt-get -t stretch-backports install "package"`

I would suggest (not mandatory, just a suggestion) that the command should use the syntax recommended by the Debian documentation.

```suggestion
 RUN apt-get update && apt-get install -t stretch-backports git-lfs
```

I assume it works "as-is" because there is no `git-lfs` package in the base Debian distribution and that causes the stretch-backports repository to be selected as the only repository that provides a `git-lfs` package.

tjanson: Neither the `apt-get install` syntax nor the git version was a deliberate choice. I’m definitely fine with following the Debian docs.

> I assume it works "as-is" because there is no `git-lfs` package in the base Debian distribution and that causes the stretch-backports repository to be selected as the only repository that provides a `git-lfs` package.

Sounds reasonable; I don’t know for sure either.

oleg-nenashev:
It should be released soon as `3.29-2`

oleg-nenashev:
@jenkinsci/java11-support I believe we need to apply the same patch to the Java 11 image

batmat:
Deemed fixed with https://github.com/jenkinsci/docker-slave/pull/63 @jenkinsci/java11-support

PR #62, opened 2019-05-21
Jenkinsfile calling simplistic build script

batmat:
Let's check things look good

Very simplistic for now.

Should help avoid https://github.com/jenkinsci/docker-slave/pull/61 / #60 in the future. cc @tjanson

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,24 @@
+pipeline {
+    agent { label 'docker' }
+
+    options {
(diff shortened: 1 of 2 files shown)


batmat:
And indeed fails as _expected_ currently and reported in #60.

batmat:
Going to merge this right away, so we can check if #61 does actually fix the issue.

Please feel free to review even after merge, I'll fix potential overlooked issues.

PR #69, opened 2019-07-12
Initial commit of Windows docker image support.

slide:
- Currently only supports Windows Server Core
- Files in place to support nanoserver, but more testing needs to be done
- Both JDK 8 and JDK 11 images are supported

Feedback welcome!

### DIFF
--- a/Dockerfile-windows
+++ b/Dockerfile-windows
@@ -1,0 +1,61 @@
+# escape=`
+
+# The MIT License
+#
+#  Copyright (c) 2019, Alex Earl and other Jenkins Contributors
+#
+#  Permission is hereby granted, free of charge, to any person obtaining a copy
+#  of this software and associated documentation files (the "Software"), to deal
+#  in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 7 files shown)


oleg-nenashev:
Plugin management is not available ATM, right?

slide:
This is the agent docker image, not the master.

oleg-nenashev:
oops

slide:
Once this is merged, how do you want to get an initial image published as we don't have agents on ci.jenkins.io (or trusted right now either) that support building docker images on Windows?

oleg-nenashev:
Manual push is fine with me (can grant you permissions), ideally would
prefer a DockerHub autobuild

slide:
DockerHub does not support Windows builds. They have had it on their roadmap since 2017, but seems no progress.

slide:
FYI, my end goal with this is to have docker agents available to ci.jenkins.io and trusted.ci. Once that is in place, these can be auto-built and published like the current Linux agent images are.

PR #71, opened 2019-07-17
Add JDK-12 Alpine-Based Image

### LINKED ISSUE #70: JDK 12 support
Is there a plan to publish a `latest-jdk12` or r a `jdk-12-alpine` version of this image?

nikeee:
This PR copies the [Dockerfile for the JDK-8 alpine image](https://github.com/jenkinsci/docker-slave/blob/master/Dockerfile-alpine) with a different base image (`openjdk:12-jdk-alpine`).

`openjdk:11-jdk-alpine` does not seem to exist. `openjdk:12-jdk` is not debian-based.

Resolves #70

### DIFF
--- a/Dockerfile-jdk12-alpine
+++ b/Dockerfile-jdk12-alpine
@@ -1,0 +1,50 @@
+# The MIT License
+#
+#  Copyright (c) 2015-2019, CloudBees, Inc. and other Jenkins contributors
+#
+#  Permission is hereby granted, free of charge, to any person obtaining a copy
+#  of this software and associated documentation files (the "Software"), to deal
+#  in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Let's keep it @nikeee . I will see what we can do about it

oleg-nenashev:
After the changes by @batmat, I would be happy to introduce new packaging destinations.
I will work on integrating it once we switch the builds to Trusted CI, CC @slide 

Maybe we would also like to switch to JDK 13 btw

nikeee:
Is this still relevant?

slide:
We don't currently have any agent images with java higher than 11. Is it needed? I think we tend to support things that are LTS within the Jenkins project.

nikeee:
To me, it's not that relevant anymore. We could close this, but I'm not aware of the goals of docker-agent.

PR #75, opened 2019-09-08
install docker to make dind

mustafakirimli:
Should we have DIND on slave to build docker images?


oleg-nenashev:
I am rather -1 w.r.t it, but needs more feedback from @jenkinsci/docker folks

slide:
I think this should be part of an inherited image rather than part of this image. This image is to he used as a base image, so you could easily make a DIND agent with this as a base. So, in short I am -1 on this.

slide:
Based on the feedback, I'm going to close this out without merging.

PR #81, opened 2019-09-20
[JENKINS-42846] - Fix the agent.jar symbolic link in the JDK 8 image

sorend:
@oleg-nenashev : Refers this: https://github.com/jenkinsci/docker-jnlp-slave/pull/100


oleg-nenashev:
Thanks! This is a race condition, because I was working on #82 with a fix for all images

PR #82, opened 2019-09-20
[JENKINS-42846] - Fix regressions in images after slave.jar => agent.jar renaming with improper symlink

oleg-nenashev:
Fixes agents. 3.35-1 is completely broken, and emergency reviews will be appreciated

### DIFF
--- a/Dockerfile-alpine
+++ b/Dockerfile-alpine
@@ -36,9 +36,10 @@ LABEL Description="This is a base image, which provides the Jenkins agent execut
 ARG AGENT_WORKDIR=/home/${user}/agent
 
 RUN apk add --update --no-cache curl bash git git-lfs openssh-client openssl procps \
-  && curl --create-dirs -fsSLo /usr/share/jenkins/slave.jar https://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/${VERSION}/remoting-${VERSION}.jar \
+  && curl --create-dirs -fsSLo /usr/share/jenkins/agent.jar https://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/${VERSION}/remoting-${VERSION}.jar \
   && chmod 755 /usr/share/jenkins \
-  && chmod 644 /usr/share/jenkins/slave.jar \
+  && chmod 644 /usr/share/jenkins/agent.jar \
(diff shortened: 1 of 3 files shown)


Dockerfile-alpine
```diff
@@ -36,9 +36,10 @@ LABEL Description="This is a base image, which provides the Jenkins agent execut
 ARG AGENT_WORKDIR=/home/${user}/agent
 
 RUN apk add --update --no-cache curl bash git git-lfs openssh-client openssl procps \
-  && curl --create-dirs -fsSLo /usr/share/jenkins/slave.jar https://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/${VERSION}/remoting-${VERSION}.jar \
+  && curl --create-dirs -fsSLo /usr/share/jenkins/agent.jar https://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/${VERSION}/remoting-${VERSION}.jar \
   && chmod 755 /usr/share/jenkins \
-  && chmod 644 /usr/share/jenkins/slave.jar \
+  && chmod 644 /usr/share/jenkins/agent.jar \
+  && ln -sf /usr/share/jenkins/agent.jar /usr/share/jenkins/slave.jar \
```
oleg-nenashev: `-f` flag here was a bad idea, yes

jimklimov: I think you don't really want full pathnames here, if you reference an `agent.jar` located in same directory as the `slave.jar` symlink. Full names make work with alternate root FSes (backups, look into containers from hosts, etc.) problematic, relative symlinks ease this pain a lot.

oleg-nenashev: I suggest doing it as a follow-up. Just fixing the images now so that they do not fall apart

jimklimov: Yes, sure, this nit is not urgent :)

stopalopa: I now can't remember why I added the -f flag, but maybe it does something different than what I originally thought. Why is it a bad idea?

oleg-nenashev:
-f flag "remove existing destination files", and hence it suppressed the
issue in this case (wrong argument ordering)

stopalopa:
ln works differently in Mac and Linux?  When I check the man pages on my Mac it says the following for -f: If the target file already exists, then unlink it so that the link may occur.  (The -f option overrides any previous -i options.)

PR #88, opened 2019-09-25
Setup Jenkinsfile to build and publish Windows container

slide:


slide:
I was trying to use the ci.j.io-jenkinsfilerunner to test it out (at least for the Linux portion), but I couldn't get things working correctly.

This may need some adjustments when actually running on trusted, but I can't really test it very well until I have this in place.

slide:
This now successfully builds on ci.j.io and publishes the docker container to jenkins4eval when built from ci.j.io.

PR #93, opened 2019-10-23
Only publish on master

slide:
Currently all PR's will publish windows images, making it only happen on master

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,10 +25,15 @@ pipeline {
                     }
                     steps {
                         script {
-                            // we can't use dockerhub builds for windows
-                            // so we publish here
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -25,10 +25,15 @@ pipeline {
                     }
                     steps {
                         script {
-                            // we can't use dockerhub builds for windows
-                            // so we publish here
-                            infra.withDockerCredentials {
-                                powershell '& ./make.ps1 publish'
+                            powershell '& ./make.ps1 build'
+                            
+                            def branchName = "${env.BRANCH_NAME}"
```
jglick:
```suggestion
                            def branchName = BRANCH_NAME
```

and inline.

PR #95, opened 2019-11-21
This adds git to windows docker images to align them more with existing linux images

Vlatombe:
@slide

### DIFF
--- a/Dockerfile-windows-nanoserver-jdk11
+++ b/Dockerfile-windows-nanoserver-jdk11
@@ -54,6 +54,15 @@ RUN [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tl
 
 ARG JAVA_HOME
 
+ARG GIT_VERSION=2.24.0
+ARG GIT_PATCH_VERSION=2
+
+RUN [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12 ; `
+    Invoke-WebRequest $('https://github.com/git-for-windows/git/releases/download/v{0}.windows.{1}/MinGit-{0}.{1}-busybox-64-bit.zip' -f $env:GIT_VERSION, $env:GIT_PATCH_VERSION) -OutFile 'mingit.zip' -UseBasicParsing ; `
+    Expand-Archive mingit.zip -DestinationPath c:\mingit ; `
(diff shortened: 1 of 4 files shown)


Dockerfile-windows-nanoserver-jdk11
```diff
@@ -54,6 +54,15 @@ RUN [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tl
...
 ARG JAVA_HOME
 
+ARG GIT_VERSION=2.24.0
+ARG GIT_PATCH_VERSION=2
+
+RUN [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12 ; `
+    Invoke-WebRequest $('https://github.com/git-for-windows/git/releases/download/v{0}.windows.{1}/MinGit-{0}.{1}-busybox-64-bit.zip' -f $env:GIT_VERSION, $env:GIT_PATCH_VERSION) -OutFile 'mingit.zip' -UseBasicParsing ; `
+    Expand-Archive mingit.zip -DestinationPath c:\mingit ; `
+    Remove-Item mingit.zip -Force ; `
+    setx /M PATH $('c:\mingit\cmd;{0}' -f $env:PATH)
```
jglick: To avoid duplication, this could be moved to a `git.ps1` file which would be `COPY`d and `RUN` by each `Dockerfile`.

Vlatombe:
> Looks good. Did you manage to test in a Kubernetes cluster in the context of a build using the git (or checkout scm) step?

No, only checked that the git command line is available in a live container.

slide:
:+1:

slide:
Just want to clarify, you tested on all four images that git is available when running the image? My concern is that nanoserver does not have some dlls, so I want to make sure that it's working for clone and other operations in those images.

Vlatombe:
Ah, the nanoserver wasn't in the makefile, I'll try it.

slide:
The nanoserver isn't in the make.ps1 yet, you are correct. We can hold off on it since its not being published yet. I can test later and update make.ps1 as needed.

Vlatombe:
I'm getting timeouts building the nanoserver image, not sure why...

slide:
Since the nanoserver images aren't published right now, let's merge this and I can look at those later.

Issue #100, opened 2020-01-11
Rework docker slave/jnlp/ssh projects to support alpine and newer openjdk versions


krufab:
At work we extensively use [docker-jnlp-slave](https://github.com/jenkinsci/docker-jnlp-slave) and [docker-ssh-slave](https://github.com/jenkinsci/docker-ssh-slave), and by extension, this docker-slave image which is the base image for docker-jnlp-slave. We mostly use the alpine version, but sometimes the stretch based one.
We found ourselves in need of using docker-jnlp-slave with the latest versions of openjdk, but it is not yet available here. Rather than working on my company's docker image only, I'd like to propose a big rework on the structure of these projects, so that it will become easier, for the future, to introduce newer images i.e. when alpine gets an upgrade or a new openjkd is out. 
I am willing to handle this rework under the supervision/approval of the repositories' owners.
I would envision the following sequence of steps:
1. Add test cases for docker-slave & docker-jnlp-slave (docker-ssh-slave has already them)
1. Rework on the projects folder structure so that there is a logical hierarchical structure which gives immediate understanding of jdk version and base image version (i.e. now it is not immediate to understand that alpine version is based on alpine3.6). This would lead to a project structure similar to the one of [node js](https://github.com/nodejs/docker-node) and [openjdk](https://github.com/docker-library/openjdk)
1. Rework on the Dockerfiles themselves to reorganize the build steps in order to achieve maximum reusability of chached layers
1. Add support for alpine (when available) and newer openjdk versions according to the openjdk releases

For each step, I'd like to work in parallel on all the projects, so that they would be all aligned.

As I would need to use the newer images at my company's work place, I would like to handle the whole process in quite short time, i.e. max 30 days, delivering each step every week or less, giving some time to the repository owners to review the changes and test them.

Please let me know your opinion, I've already some code done and I could propose PRs early next week.

krufab:
@oleg-nenashev

MarkEWaite:
@krufab I am very interested in the technique you're proposing and the structure you're proposing.  I think you should begin the work and we'll do our best to keep pace with code review of your proposed changes.

[...]

PR #101, opened 2020-01-17
Added tests to the project

krufab:
Added tests to the project, as per issue #100


slide:
The problem is that you aren't a trusted contributor, so the CI will not use your Jenkinsfile changes.

tests/tests.bats
```diff
@@ -0,0 +1,122 @@
...
+  [ "${JDK}" = "${lines[0]}" ]
+
+  run docker exec "${SLAVE_CONTAINER}" sh -c "printenv | grep AGENT_WORKDIR"
+  [ "AGENT_WORKDIR=/home/jenkins/agent" = "${output}" ]
+}
+
+@test "[${FLAVOR}] use build args correctly" {
+  cd "${BATS_TEST_DIRNAME}"/.. || false
+
+  local TEST_VERSION="3.36"
```
oleg-nenashev: Would be nice to avoid putting it here directly. It will become immediately obsolete otherwise (now we are on Remoting 3.40)

tests/tests.bats
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env bats
+
+DOCKERFILE=Dockerfile
+JDK=8
+SLAVE_IMAGE=jenkins-slave
```
oleg-nenashev: Please use "AGENT" instead of "SLAVE" for new code

tests/tests.bats
krufab: Ok, but that is a test: I'd like to be sure that the image allows the specification of a some different version than the one hardcoded in the Dockerfile ARG VERSION=3.40

krufab:
@oleg-nenashev, there is an integration test in jnlp-slave.
I agree integration tests would be useful as well. However they might come later on...

@slide , i'll revert the build.sh so that the tests can be executed

slide:
@krufab I was not saying that you should revert build.sh, I am just telling you why the CI is failing. What I can do is submit a shadow PR with your same changes so that the CI will actually run using the Jenkinsfile changes, would that be ok with you?

krufab:
@slide, I guess it's ok, yes

slide:
@krufab See #102

tests/tests.bats
krufab: Ok, done

krufab:
@slide, Bats doesn't seem to be installed on the test server. I reverted the build.sh file, removed make tests and now it builds. You can run the tests on your local machine and check if they are ok. When the build server will be able to run the tests, I'll re-enable them.

slide:
If you look at the master docker repo, you can see that bats is a submodule. That would be the way to go here too.

build.sh
```diff
@@ -1,20 +1,3 @@
...
-    if [[ "$dockertag" = "$dockerfile" ]]; then
-        dockertag='latest'
-    fi
-    echo "Building $dockerfile => tag=$dockertag"
-    docker build -f $dockerfile -t jenkins/slave:$dockertag .
-    docker build -f $dockerfile -t jenkins/agent:$dockertag .
-done
-
-echo "Build finished successfully"
+make build
```
slide: This should also `make tests`

slide:
I pulled in all the latest changes into my duplicate PR so we can see the Jenkinsfile changes in action

build.sh
krufab: Done.

slide:
#102 is passing for all the checks including the tests

krufab:
Is it ok this PR or is there anything to change?

slide:
I am happy with this, if @oleg-nenashev has no objections, I think it can be merged. I will look at adding some simple pester based tests for the Windows side.

PR #106, opened 2020-02-17
Provide image which is based on Debian 10/buster

krichter722:
Debian 9 although stable is quite old in terms of up-to-date versions of software.

### DIFF
--- a/Dockerfile-jdk11-buster
+++ b/Dockerfile-jdk11-buster
@@ -1,0 +1,50 @@
+# The MIT License
+#
+#  Copyright (c) 2015-2020, CloudBees, Inc. and other Jenkins contributors
+#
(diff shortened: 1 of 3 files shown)


krichter722:
@rtyler Thanks for merging this. Can you estimate when the image tag `latest-jdk11-buster` will be available?

slide:
I think the dockerhub builds need to be updated to.build that dockerfile

krichter722:
@slide Can this be done in a PR that I could create or is it an administrative task that needs to be done by someone with access rights.

slide:
It needs to be done by someone with write access on the repo on dockerhub. I am not sure who that is.

slide:
You can see the existing builds here: https://hub.docker.com/r/jenkins/slave/builds

One would need to be added to create a build using the dockerfile added.

PR #110, opened 2020-03-25
Update to Remoting 4.3 for WebSocket large payload fix.

jeffret-b:
See [JENKINS-61409](https://issues.jenkins-ci.org/browse/JENKINS-61409).

Also see the [Remoting changelog](https://github.com/jenkinsci/remoting/releases/tag/remoting-4.3).

Related to jenkinsci/jenkins#4601.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@
 FROM openjdk:8-jdk
 MAINTAINER Oleg Nenashev <o.v.nenashev@gmail.com>
 
-ARG VERSION=4.0.1
+ARG VERSION=4.3
 ARG user=jenkins
 ARG group=jenkins
 ARG uid=1000
--- a/Dockerfile-alpine
+++ b/Dockerfile-alpine
@@ -23,7 +23,7 @@
 FROM openjdk:8-jdk-alpine
 MAINTAINER Oleg Nenashev <o.v.nenashev@gmail.com>
 
-ARG VERSION=4.0.1
+ARG VERSION=4.3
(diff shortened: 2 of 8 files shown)


jeffret-b:
Jesse has a great idea. Do you or someone on the team want to implement that? Or I could see about putting something together.

jeffret-b:
See if we can get clean rebuild.

oleg-nenashev:
> Jesse has a great idea. Do you or someone on the team want to implement that? Or I could see about putting something together.

We still use the DockerHub Autobuild as a delivery engine. One we move to the Jenkins release environment, we can change the implementation

jeffret-b:
Looks like some timeout in the build.

MarkEWaite:
It is looking for an agent with the label `windock`.  There is only one agent with that label currently.  It is https://ci.jenkins.io/computer/win2019-a7d1c0/ and it is offline

@slide I'm not clear on the ci.jenkins.io agent labels for Windows machines.  I assume `windock` means a Windows machine that has Docker for Windows installed.  Is it also assumed that label means the Windows machine can run Windows Docker images (as opposed to Linux images)?

slide:
@MarkEWaite, yes there are two different Windows images because most plugins assume Linux when running on an agent with the docker label, also the image is larger when it has docker installed because there is no Windows Server Core sku with docker on Azure.

slide:
It's only been running for 7 minutes so far, the Windows agents can take some time to spin up, so this time doesn't see out of the ordinary.

Issue #111, opened 2020-03-26
The password of jenkins user in Windows images expires after a few weeks


sebok:
Derived windows images became unusable after the (empty) password of the jenkins user expires.
At this time jenkins/jnlp-agent:latest-windows fails to start, derived images can not be built.

**Workaround:**
Rebuild all base images locally.

slide:
This is fixed the latest release.

PR #112, opened 2020-03-26
Let jenkins user's password in windows never expire.

sebok:
See issue #111 The password of jenkins user in Windows images expires after a few weeks.


slide:
I have a lot of updates I am preparing to create a PR for that will fix this. I will look at this and see how it compares to the solution I devised.

PR #120, opened 2020-04-10
Cleanup "slave.jar" references in Java 11 Debian image descriptions

oleg-nenashev:


oleg-nenashev:
I am not sure why Docker Ci fails, looks like something is broken in their caching. Will merge as is

PR #126, opened 2020-04-13
Stop shipping undocumented Windows image tags

### LINKED ISSUE #125: Collision between Linux and Windows flows in the `latest` tag
I have missed it while preparing the 4.3-2 release...

Apparently Windows flow also deploys the `latest` image, and now there is eventual  consistency in this tag.

Fineally one ends up with Windows image

```
f2ee63f0fe3c: Waiting                                                                                    675c1c1b424b: Waiting                                                                                    135a276a8e43: Waiting                                                                                    docker: image operating system "windows" cannot be used on this platform.
See 'docker run --help'.
```

oleg-nenashev:
This change restricts deployments only to tags documented in https://github.com/jenkinsci/docker-agent#configurations . We should definitely start shipping short handles for Windows images later, but now it is needed to prevent collision of the `latest` tag.

Closes #125

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -23,11 +23,11 @@ if(![String]::IsNullOrWhiteSpace($env:DOCKERHUB_ORGANISATION)) {
 $builds = @{
     'jdk8' = @{
         'Folder' = '8\windows\windowsservercore-1809';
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Will merge it to fix the release, will follow-up later this week

oleg-nenashev:
Linux build times out, but it should not be any worse

PR #133, opened 2020-04-29
Update directory structure and switch to AdoptOpenJDK where possible

slide:
Updated directory structure for Debian images to group
them into a top level directory.

Updated some images to use AdoptOpenJDK where possible.


slide:
I think I may want to split this PR up into a couple, but would like to know what others think about that.

oleg-nenashev:
In the Platform SIG we already agreed to move to AdoptOpenJDK by default. My suggestion would be to keep OpenJDK images as tags for compatibility reasons (users of latest will likely be impacted anyway)

slide:
We can't move all images to Adopt, they don't provide Debian images for stretch from what I could see.

slide:
I am going to split this up into AdoptOpenJDK image change and other stuff.

PR #135, opened 2020-05-09
README: Add explicit tag to the Windows example

oleg-nenashev:


oleg-nenashev:
Ec2 provisioning fails. I will merge it, because there is nothing really to be tested

PR #142, opened 2020-07-06
Update to Remoting 4.5.

jeffret-b:
See jenkinsci/jenkins#4831

With the 4.5 release, the Remoting library is now released on the Jenkins release system and signed with the CDF certificate.

The 4.4 and 4.5 releases involve minor maintenance changes to improve ongoing maintainability. See the Remoting changelogs: [4.4](https://github.com/jenkinsci/remoting/releases/tag/remoting-4.4) and [4.5](https://github.com/jenkinsci/remoting/releases/tag/remoting-4.5).

PR #143, opened 2020-07-15
Simplify updating the Remoting version

jeffret-b:
This reduces to two the number of places where the current Remoting version needs to be updated, once for Windows, and once for Linux. Recent changes distributing Dockerfiles into subdirectories made it even more difficult to update each of the locations, which was already a hassle.

I'd like to reduce the number of places to one, but I'm not sure how we would want to do that. Put something in an environment file or something?

I don't know what to do about the tests. They're already very outdated as they build their own images based on 3.36. I would think they should be updated, but I don't understand why they're already so far behind. Or why they don't test what is built.


jeffret-b:
Sounds like it's okay to leave them as-is, then.

slide:
We would definitely use an environment file I think. Just something that specifies a list of variables to specify on the build?

```
VERSION=4.45
ARGFOO=bar
```

This should be parseable by both bash and powershell.

jeffret-b:
Something like that seems like an even better improvement over this. Is there standard parsing for this in existing tooling? Bash, PS ...

slide:
for bash you can do something like this:

with a file named env.props
```
VERSION=4.3
```

```
source env.props
export (cut -d= -f1 env.props)
```

VERSION would now be an env variable, something similar could be done to generate an "extra_args" variable to pass to the docker build commands.

For powershell, I would have to look into it a little more, but it should be pretty easy to parse the file into a hashtable or something similar and then create an extra_args variable to pass.

jeffret-b:
I'll try updating this PR with the suggested change for bash. I don't have PS very handy.

jeffret-b:
I integrated a new env.props file for the Linux side. Looks like it works fine.

slide:
I added it for the windows side and tested locally and it seems to work fine there as well.

jeffret-b:
Thanks for the commit! That looks good. This is what I was wanting to get to.

slide:
One thing I just thought about though is the case where we are releasing a new version of this image via a tag. We may need to update the Jenkinsfile, or change the logic in make.ps1 for this case where the RemotingVersion is specified on the command line. We could change the Jenkinsfile to set an environment parameter (REMOTING_VERSION) instead of using the command line parameter. Thoughts?

jeffret-b:
I was wondering about other scenarios like that. I'm not sure how that would work, though. Are you suggesting putting the REMOTING_VERSION in the Jenkinsfile? That doesn't sound like a good idea.

We could put the code to read env.props behind a check to see if an environment variable is already set. If the environment variable doesn't exist then read the file and set the environment variable from that.

slide:
What I mean would be that we would do something like only for the tag builds. 

```
if(env.TAG_NAME != null) {
                                def tagItems = env.TAG_NAME.split('-')
                                if(tagItems.length == 2) {
                                    def remotingVersion = tagItems[0]
                                    def buildNumber = tagItems[1]
                                    // we need to build and publish the tag version
                                    withEnv(["REMOTING_VERSION=${remotingVersion}"]) {
                                        infra.withDockerCredentials {
                                            powershell "& ./make.ps1 -PushVersions -BuildNumber $buildNumber publish"
                                        }
                                    }
                                }
                            }
```

We would still need to check if the env variable is already set before picking it up from the env.props file. 

This shouldn't matter for Linux images because those are built on dockerhub. Which, now that I think about it, would mean that the individual dockerfiles would need to be updated anyway...so this simplification may not work at all.

jeffret-b:
> Which, now that I think about it, would mean that the individual dockerfiles would need to be updated anyway...so this simplification may not work at all.

Uggh.

That's annoying.

We've got to do something about this version indicator spread throughout directories and files. If we can't script the usage, at least we need to create a utility to update all of the instances.

slide:
I think a utility would be a good idea, I can take a crack at a python script or something

jeffret-b:
@slide, are you going to try to put together a script for this? Otherwise I can see what I can do when I'm back from vacation in a couple of weeks.

slide:
I'm actually planning on creating a PR that moves the builds into our infra so we can control some things more than we can on dockerhub. This would allow us to use the env file with the version. I do have a simple script that updates the various files that I can share as well if desired.

jeffret-b:
I'm fine with whatever approach you'd like to take. That one sounds like a good way to go. If you'd like to take over this PR as part of that, especially while I'm out, please do. There are also the downstream PRs to this: #142 and jenkinsci/docker-inbound-agent#182. When I'm back I can see how I can help out.

jeffret-b:
@slide, while there's no significant rush, it would be nice to update the images to the 4.5 Remoting version to complete that effort. Do you think it would be best if I just update them manually for now? Or should we keep working towards a simplified, maintainable solution?

slide:
Yes, let's do it manually for now. I hope to have time next week to do PR to publish all images from our infra so we can make it easier on ourselves

jeffret-b:
If you think you'd have something ready for next week, I'd be happy to wait until then. We kind of need a reason to update in order to run this sort of simplification anyway. If it would take us another month, I'd be more interested in going manual. If it takes a week or two, I think it would be worthwhile to make sure we get it working right on your changes. Thoughts?

MarkEWaite:
@jeffret-b I resolved the conflict to avoid a recently discovered [git plugin bug](https://issues.jenkins-ci.org/browse/JENKINS-63751).  My apologies if the conflict resolution is incorrect, but I wanted to stop the hourly builds of this pull request on ci.jenkins.io without requiring that I spend the hours necessary to debug the problem in the git plugin.

PR #144, opened 2020-07-15
Fix issue with Windows passwords expiring

slide:
Uses wmic on servercore and sets no expiration on nanoserver


oleg-nenashev:
Addresses https://github.com/jenkinsci/docker-inbound-agent/issues/183 and https://github.com/jenkinsci/docker-inbound-agent/issues/173

PR #152, opened 2020-08-13
Rework publishing to be completely in the Jenkins infra

slide:
Reworks the build.sh script to mimick the build.ps1 script to provide the ability to build different images and push them withing the Jenkins infra (ci.j.io and trusted.ci). We would be able to manage the versions being built by just updating the env.props file with the new remoting version.

This is a follow on to the discussion in https://github.com/jenkinsci/docker-agent/pull/143


Jenkinsfile
```diff
@@ -24,15 +24,14 @@ pipeline {
                         DOCKERHUB_ORGANISATION = "${infra.isTrusted() ? 'jenkins' : 'jenkins4eval'}"
                     }
                     steps {
+                        powershell '& ./build.ps1 test'
                         script {
-                            powershell '& ./make.ps1 test'
-
                             def branchName = "${env.BRANCH_NAME}"
                             if (branchName ==~ 'master') {
                                 // we can't use dockerhub builds for windows
```
jeffret-b, line 31: I don't think this comment is relevant any more.

Jenkinsfile
```diff
@@ -60,12 +58,35 @@ pipeline {
...
+                    }
                     steps {
+                        sh './build.sh'
                         script {
-                            if(!infra.isTrusted()) {
-                                sh './build.sh ; docker system prune --force --all'
+                            def branchName = "${env.BRANCH_NAME}"
+                            if (branchName ==~ 'master') {
+                                // we can't use dockerhub builds for windows
+                                // so we publish here
```
jeffret-b, line 70: A copy of a comment that probably isn't relevant anymore.

jeffret-b:
Any chance we can get something moving on this?

I really ought to get the latest Remoting into these docker versions. I'd rather do it this way, but could do it more manually if needed.

jeffret-b:
@slide, where are we with any of this? I don't think we ever integrated the Remoting 4.5 version into these docker images. I've just released the 4.6 version and want to try again to get that integrated. Should I just use a brute-force mechanism?

slide:
Yes, brute force for now sadly...

oleg-nenashev (approved):
I do not see anything wrong here. The current DockerHub flow is broken, so it definitely makes sense to get this PR finalized rather sooner than later. Thanks @slide !

MRamonLeon (approved):

Issue #156, opened 2020-10-31
Test failures don't fail the build in Windows


v1v:
### Stacktrace

```
[2020-10-31T15:58:58.899Z] stderr:
[2020-10-31T15:58:58.899Z] Template parsing error: template: :1: unclosed action
[2020-10-31T15:58:58.899Z] 
[2020-10-31T15:58:58.899Z] 
[2020-10-31T15:58:58.899Z] 
[2020-10-31T15:58:58.899Z]     [-] version in docker metadata 136ms
[2020-10-31T15:58:58.899Z]       Expected 0, but got 64.
[2020-10-31T15:58:58.899Z]       191:         $exitCode | Should -Be 0
[2020-10-31T15:58:58.899Z]       at <ScriptBlock>, C:\Jenkins\workspace\Packaging_docker-agent_PR-141\tests\agent.Tests.ps1: line 191
[2020-10-31T15:58:59.648Z] Tests completed in 102.95s
[2020-10-31T15:58:59.648Z] Tests Passed: 15, Failed: 1, Skipped: 0, Pending: 0, Inconclusive: 0 
[2020-10-31T15:58:59.648Z] Build finished successfully
```

Build [logs](https://ci.jenkins.io/blue/organizations/jenkins/Packaging%2Fdocker-agent/detail/PR-141/28/pipeline#step-47-log-1229)

PR #158, opened 2020-10-31
[Hactoberfest] Fail the execution at the end if any test failures

v1v:
Closes https://github.com/jenkinsci/docker-agent/issues/156

### What

Fail the build if any test failures happened when running the windows tests.

It does not fail fast to be able to provide as many details as possible.

### Why

It does not fail if test failures at the moment

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -85,7 +85,7 @@ if($lastExitCode -ne 0) {
 
 if($target -eq "test") {
     # Only fail the run afterwards in case of any test failures
-    $testFailed = 0
+    $testFailed = $false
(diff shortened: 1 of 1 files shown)


slide:
This looks good to me, did you want to look at it more and thus changed it to draft?

slide (approved):

v1v:
I'm running some more tests to validate it works as expected, for such I'm using https://github.com/jenkinsci/docker/pull/1011

v1v:
This is now ready to be reviewed :)

PR #163, opened 2020-12-11
feat: add windows / jdk8 images for LTSC 2019

garethjevans:


slide:
Similar question as to docker-ssh-agent, do we need jdk8 versions?

PR #173, opened 2021-05-26
Update upstream images

slide:
Updates to newer upstream versions and removes stretch images.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!


daniel-beck:
@timja Do you know how we could set up Dependabot to help with parent image updates here? In particular https://github.com/jenkinsci/docker-agent/pull/173/files#diff-803a39ce4625d71616af984036ad44f73040f19b5fe9a73318980cb48207d412R26 looks feasible if it's possible to define patterns or something like that.

timja:
> @timja Do you know how we could set up Dependabot to help with parent image updates here? In particular [#173 (files)](https://github.com/jenkinsci/docker-agent/pull/173/files#diff-803a39ce4625d71616af984036ad44f73040f19b5fe9a73318980cb48207d412R26) looks feasible if it's possible to define patterns or something like that.

Should be able to, as far as I can tell dependabot just tries to be smart with image tags, so it depends on if its smart enough to handle the formats.

We use dependabot for updating jenkins core versions at work and it's fine, but that's a very simple format (https://github.com/hmcts/cnp-jenkins-docker/blob/master/.dependabot/config.yml#L3-L5 still on v1)

PR #186, opened 2021-08-19
Switch to eclipse temurin, build via jlink, switch to bullseye

timja:
See https://github.com/jenkinsci/docker/pull/1185

### DIFF
--- a/11/archlinux/Dockerfile
+++ b/11/archlinux/Dockerfile
@@ -20,7 +20,22 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
-FROM archlinux:latest AS archlinux
+FROM archlinux:latest AS jre-build
+
+RUN pacman -Syu jdk11-openjdk --noconfirm
+
+# Generate smaller java runtime without unneeded files
+# for now we include the full module path to maintain compatibility
+# while still saving space (approx 200mb from the full distribution)
+RUN /usr/lib/jvm/default-runtime/bin/jlink \
+         --add-modules ALL-MODULE-PATH \
+         --strip-debug \
+         --no-man-pages \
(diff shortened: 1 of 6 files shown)


11/archlinux/Dockerfile
```diff
@@ -20,7 +20,22 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
-FROM archlinux:latest AS archlinux
+FROM archlinux:latest AS jre-build
```
MarkEWaite, line 23: This is intentionally not coming from Eclipse Temurin?
```suggestion
FROM archlinux:latest AS jre-build
```

MarkEWaite (commented):
One question on the archlinux change

11/archlinux/Dockerfile
timja: I didn’t bother no, I question whether this image should be here at all.

I wasn’t sure if there was implications from getting it from a different OS type like Debian whether libraries would be different…

MarkEWaite (approved):

dduportal (approved):
🚀 Good work, sounds really good! (the glibc part on Alpine is tricky but needed until eclipse-temurin provides their own Alpine - https://github.com/adoptium/temurin-build/projects/1?fullscreen=true&card_filter_query=label%3Aalpine-linux)

MarkEWaite:
> 🚀 Good work, sounds really good! (the glibc part on Alpine is tricky but needed until eclipse-temurin provides their own Alpine - https://github.com/adoptium/temurin-build/projects/1?fullscreen=true&card_filter_query=label%3Aalpine-linux)

Eclipse has provided their own alpine as an unofficial image.  We used it previously.  However, I like the transition to Eclipse Temurin even more than using unofficial Adoptium images.

PR #191, opened 2021-09-02
Correct locale

timja:


MarkEWaite (approved):
I don't understand the CI failures on Windows.  They should be unaffected by this change.

timja:
> I don't understand the CI failures on Windows. They should be unaffected by this change.

it might be a symptom of rate limiting, they have passed on this PR previously with no change so I plan to just ignore it

PR #193, opened 2021-09-03
Use alpine's JRE instead of AdoptOpenJDK

username-is-too-long:
[AdoptOpenJDK](https://github.com/AdoptOpenJDK/openjdk-docker) now has a DEPRECATION NOTICE warning, and Alpine has its own packaged openjdk/jre that uses musl libc, so lines 23-35 and 55-86 could be replaced.


MarkEWaite:
I recommend against this change based on the comments in jenkinsci/docker#1185 and in jenkinsci/docker#1186.  I'm open to reconsider based on guidance from @timja and @dduportal , but the transition to Eclipse Temurin (AdoptOpenJDK) has been a positive change.  One Java provider for all our Docker images with synchronized updates across all our platforms (AMD64, Arm64, PowerPC, and s390x).

timja:
I’ll look at this closer later but I think it might be better to take this as a temporary switch until temurin provides a musl based build,

see https://github.com/adoptium/containers/issues/1 for background

username-is-too-long:
> see [adoptium/containers#1](https://github.com/adoptium/containers/issues/1) for background

Oh nice, I didn't see they were moving to musl.

timja:
> > see [adoptium/containers#1](https://github.com/adoptium/containers/issues/1) for background
> 
> Oh nice, I didn't see they were moving to musl on Alpine.

It looks like awhile away, they first need to TCK 16+ and then someone needs to backport to 11 and 8. They also need to setup some infrastructure for it.


I wouldn’t expect it to be quick but I think we can swap to this in the meantime

timja:
@username-is-too-long the locale appears to not be set to a utf-8 one can you check it?

there was some code in the glibc section doing that, it might just be a matter of switching the locale line to:

```
ENV LANG C.UTF-8
```

but I don't think that worked last time I tried, see:
https://grrr.tech/posts/2020/add-locales-to-alpine-linux-docker-image/

username-is-too-long:
> @username-is-too-long the locale appears to not be set to a utf-8 one can you check it?
> 
> there was some code in the glibc section doing that, it might just be a matter of switching the locale line to:
> 

`/usr/glibc-compat/bin/locale` doesn't exist without without glibc, I added musl-locales to provide `locale`, and removed that alpine case in tests.bats.

11/alpine/Dockerfile
```diff
@@ -49,54 +36,14 @@ ARG AGENT_WORKDIR=/home/${user}/agent
...
-    && curl -LfsS ${ZLIB_URL} -o /tmp/libz.tar.xz \
-    && echo "${ZLIB_SHA256} */tmp/libz.tar.xz" | sha256sum -c - \
-    && mkdir /tmp/libz \
-    && tar -xf /tmp/libz.tar.xz -C /tmp/libz \
-    && mv /tmp/libz/usr/lib/libz.so* /usr/glibc-compat/lib \
-    && apk del --purge .build-deps glibc-i18n \
-    && rm -rf /tmp/*.apk /tmp/gcc /tmp/gcc-libs.tar* /tmp/libz /tmp/libz.tar.xz /var/cache/apk/*
-
-RUN apk add --update --no-cache curl bash git git-lfs openssh-client openssl procps \
+RUN apk add --update --no-cache curl bash git git-lfs musl-locales openjdk11-jre-headless openssh-client openssl procps \
```
timja, line 39: we should still be running `jlink` to get a trimmed down runtime, it should be smaller than the `jre`

timja:
(I sorted the build separately in a different PR, green now 😄 )

11/alpine/Dockerfile
username-is-too-long: You're right, saved about 80mb

11/alpine/Dockerfile
```diff
@@ -20,12 +20,15 @@
...
 
-FROM adoptopenjdk/openjdk11:jdk-11.0.12_7-alpine AS jre-build
+FROM alpine:3.14.1 as jre-build
+
+RUN apk add --update --no-cache openjdk11-jdk openjdk11-jmods
 
 # Generate smaller java runtime without unneeded files
 # for now we include the full module path to maintain compatibility
-# while still saving space (approx 200mb from the full distribution)
+# while still saving space (approx 200mb from the full distribution)
```
timja, line 29:
```suggestion
# while still saving space (approx 200mb from the full distribution)
```

11/alpine/Dockerfile
```diff
@@ -20,12 +20,15 @@
...
+FROM alpine:3.14.1 as jre-build
+
+RUN apk add --update --no-cache openjdk11-jdk openjdk11-jmods
 
 # Generate smaller java runtime without unneeded files
 # for now we include the full module path to maintain compatibility
-# while still saving space (approx 200mb from the full distribution)
+# while still saving space (approx 200mb from the full distribution)                                 
 RUN jlink \
+	 --module-path /usr/lib/jvm/default-jvm/jmods \
```
timja, line 31:
```suggestion
             --module-path /usr/lib/jvm/default-jvm/jmods \
```

11/alpine/Dockerfile
```diff
@@ -49,51 +52,15 @@ ARG AGENT_WORKDIR=/home/${user}/agent
...
+RUN apk add --update --no-cache curl bash git git-lfs musl-locales openssh-client openssl procps \
   && curl --create-dirs -fsSLo /usr/share/jenkins/agent.jar https://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/${VERSION}/remoting-${VERSION}.jar \
   && chmod 755 /usr/share/jenkins \
   && chmod 644 /usr/share/jenkins/agent.jar \
   && ln -sf /usr/share/jenkins/agent.jar /usr/share/jenkins/slave.jar \
   && apk del --purge curl \
   && rm -rf /tmp/*.apk /tmp/gcc /tmp/gcc-libs.tar* /tmp/libz /tmp/libz.tar.xz /var/cache/apk/*
 
-ENV JAVA_HOME=/opt/java/openjdk
+ENV JAVA_HOME=/opt/java/openjdk
```
timja, line 63:
```suggestion
ENV JAVA_HOME=/opt/java/openjdk        
```

8/alpine/Dockerfile
```diff
@@ -20,10 +20,7 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
-# FIXME(oleg_nenashev): This is not an official AdoptOpenJDK Docker Image.
-# There is no official Alpine images at the moment.
-# Needs upgrade when/if there is an official alpine image.
-FROM adoptopenjdk/openjdk8:jdk8u302-b08-alpine
+FROM alpine:3.14.1
```
timja, line 23:
```suggestion
FROM alpine:3.14.2
```

11/alpine/Dockerfile
```diff
@@ -20,12 +20,15 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
-FROM adoptopenjdk/openjdk11:jdk-11.0.12_7-alpine AS jre-build
+FROM alpine:3.14.1 as jre-build
```
timja, line 23:
```suggestion
FROM alpine:3.14.2 as jre-build
```

11/alpine/Dockerfile
```diff
@@ -20,12 +20,15 @@
...
-FROM adoptopenjdk/openjdk11:jdk-11.0.12_7-alpine AS jre-build
+FROM alpine:3.14.2 as jre-build
+
+RUN apk add --update --no-cache openjdk11-jdk openjdk11-jmods
 
 # Generate smaller java runtime without unneeded files
 # for now we include the full module path to maintain compatibility
 # while still saving space (approx 200mb from the full distribution)
 RUN jlink \
+             --module-path /usr/lib/jvm/default-jvm/jmods \
```
timja, line 31:
```suggestion
         --module-path /usr/lib/jvm/default-jvm/jmods \
```

timja (approved):

PR #194, opened 2021-09-04
Integrate bats-assert + fix SC issues in test

timja:
[bats-assert](https://github.com/bats-core/bats-assert) will give us the actual error when tests fail, unlike in https://github.com/jenkinsci/docker-agent/pull/193 where the locale is broken but it just says test failed.

we already use it in https://github.com/jenkinsci/docker

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "tests/test_helper/bats-assert"]
+	path = tests/test_helper/bats-assert
+	url = https://github.com/bats-core/bats-assert.git
+[submodule "tests/test_helper/bats-support"]
+	path = tests/test_helper/bats-support
+	url = https://github.com/bats-core/bats-support.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "tests/test_helper/bats-assert"]
 	path = tests/test_helper/bats-assert
-	url = https://github.com/bats-core/bats-assert.git
+	url = https://github.com/bats-core/bats-assert
 [submodule "tests/test_helper/bats-support"]
 	path = tests/test_helper/bats-support
-	url = https://github.com/bats-core/bats-support.git
+	url = https://github.com/bats-core/bats-support
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "tests/test_helper/bats-assert"]
 	path = tests/test_helper/bats-assert
-	url = https://github.com/bats-core/bats-assert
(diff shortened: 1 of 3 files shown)


tests/tests.bats
```diff
@@ -1,16 +1,17 @@
 #!/usr/bin/env bats
 
 load test_helpers
+load 'test_helper/bats-support/load' # this is required by bats-assert!
+load 'test_helper/bats-assert/load'
 
-REGEX='^([0-9]+)/(.+)$'
-
+IMAGE=${IMAGE:-debian_jdk11}
```
timja, line 7: supply a default value for when running outside of `make`, (like in IDE)

tests/tests.bats
```diff
@@ -1,16 +1,17 @@
 #!/usr/bin/env bats
 
 load test_helpers
+load 'test_helper/bats-support/load' # this is required by bats-assert!
+load 'test_helper/bats-assert/load'
 
-REGEX='^([0-9]+)/(.+)$'
```
timja, line 5: unused variable

MarkEWaite (approved):
Looks good to me, though I agree that the two comments might be nice to include before merge

timja:
> Looks good to me, though I agree that the two comments might be nice to include before merge

2 comments?

MarkEWaite:
> > Looks good to me, though I agree that the two comments might be nice to include before merge
> 
> 2 comments?

https://github.com/jenkinsci/docker-agent/pull/194#discussion_r702288421 and 

> > Looks good to me, though I agree that the two comments might be nice to include before merge
> 
> 2 comments?

I missed that your review comments were not proposing any change, just explaining the rationale for the change in that line

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "tests/test_helper/bats-assert"]
+	path = tests/test_helper/bats-assert
+	url = https://github.com/bats-core/bats-assert.git
```
timja, line 3:
```suggestion
	url = https://github.com/bats-core/bats-assert
```

.gitmodules
```diff
@@ -0,0 +1,6 @@
...
+[submodule "tests/test_helper/bats-support"]
+	path = tests/test_helper/bats-support
+	url = https://github.com/bats-core/bats-support.git
```
timja, line 6:
```suggestion
	url = https://github.com/bats-core/bats-support
```

timja:
Created a new PR due to issues on CI, ref https://newbedev.com/git-clone-works-git-submodule-fails-permission-denied

PR #195, opened 2021-09-04
Integrate bats-assert + fix SC issues in test

timja:
Same as https://github.com/jenkinsci/docker-agent/pull/194

.git/config doesn't seem to be updating due to https://newbedev.com/git-clone-works-git-submodule-fails-permission-denied

[bats-assert](https://github.com/bats-core/bats-assert) will give us the actual error when tests fail, unlike in https://github.com/jenkinsci/docker-agent/pull/193 where the locale is broken but it just says test failed.

we already use it in https://github.com/jenkinsci/docker

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "tests/test_helper/bats-assert"]
+	path = tests/test_helper/bats-assert
+	url = https://github.com/bats-core/bats-assert.git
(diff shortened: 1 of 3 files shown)


timja:
Approved in https://github.com/jenkinsci/docker-agent/pull/194, will look at windows build separately

PR #196, opened 2021-09-06
Update windows images to eclipse temurin

timja:
Also pretend to be curl in user agent so that fetching remoting works (User agent sniffing is happening -.-)

ltsc2019 and 1809 are just aliases of each other apparently: https://github.com/adoptium/containers/issues/35


MarkEWaite (approved):
Looks like the right thing to me.  Learned something new today by reading about the intended use of MinGit.  I always use PortableGit on my Windows computer, but MinGit looks like the perfect solution for CI agents.  Thanks

timja:
> Looks like the right thing to me. Learned something new today by reading about the intended use of MinGit. I always use PortableGit on my Windows computer, but MinGit looks like the perfect solution for CI agents. Thanks

ha I have no idea about it, just seemed like a good idea to update while I was here

PR #201, opened 2021-09-23
Add java 17 preview

timja:
Build will fail till https://github.com/docker-library/official-images/pull/10945 is merged

### DIFF
--- a/17/bullseye/Dockerfile
+++ b/17/bullseye/Dockerfile
@@ -1,6 +1,6 @@
 # The MIT License
 #
-#  Copyright (c) 2015-2020, CloudBees, Inc. and other Jenkins contributors
+#  Copyright (c) 2015-2021, CloudBees, Inc. and other Jenkins contributors
 #
 #  Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


11/bullseye/Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN jlink \
          --compress=2 \
          --output /javaruntime
 
-FROM debian:bullseye-20210816
+FROM debian:bullseye
```
MarkEWaite, line 36: I prefer the date stamp on the Docker image so that we track how close we are to the current date.  Is there a compelling reason why this needs to remove the date code?
```suggestion
FROM debian:bullseye-20210816
```

MarkEWaite (approved):
Approved but with a question on the change to a less precise debian image version

11/bullseye/Dockerfile
timja: dependabot doesn't work with it and we don't have enough maintainers.
We have a higher chance of staying up to date if we don't fix it to this tag.

but if you feel strongly I can change back

timja: can do a follow up if required

PR #228, opened 2022-01-14
Added alpine tags with OS version (i.e. alpine3.15)

### LINKED ISSUE #227: Add tags with the base image version (i.e. alpine3.15-jdk11)
### What feature do you want to see added?

Please, do add a tag containing the version of the base image used (i.e. for alpine, archlinux) otherwise dependent images will not be able to distinguish when you move from alpine3.14 to alpine3.15, leading to a serous incompatibilities.

For example, docker inbound-agent uses: `ARG version=latest-alpine-jdk11` https://github.com/jenkinsci/docker-inbound-agent/blob/7f45726f463c68e9321886f0ff1e2e65bfea44a6/11/alpine/Dockerfile#L24 (we can agree that using latest is quite wrong, but there is no tag for docker agent alpine3.14, to stick to that one: https://hub.docker.com/r/jenkins/agent/tags?page=1&name=alpine3.14 )

Today I rebuild my custom image based on inbound-agent and i could not connect anymore to the old servers as now the base image is alpine 3.15, with updated openssh libraries (8.8), which has disabled the `ssh-rsa` HostKeyAlgorithm and I cannot connect to old linux servers... :/

I'd suggest something like:
- latest-alpine-jdk11, alpine3.15-jdk11, alpine-jdk11 etc to be tags for one version
- alpine3.14-jdk11 to be the tags of the old alpine version

so that when you'll move to alpine3.16, then alpine3.15-jdk11 will still be available

A similar approach was suggested here: https://github.com/jenkinsci/docker-agent/pull/105#issuecomment-580144908

For whoever is affected: a workaround for this issue is to add:
```
RUN set -e; \
  echo -e "Host * \n  HostkeyAlgorithms +ssh-rsa\n  PubkeyAcceptedAlgorithms +ssh-rsa" >> /etc/ssh/ssh_config
```
to the Dockerfile

### Upstream changes

_No response_

krufab:
Signed-off-by: Fabio Kruger <10956489+krufab@users.noreply.github.com>

Updated docker-back.hcl to create new tags for the alpine target containing the alpine OS release as well (i.e. alpine3.15-jdk11).
Archlinux was not changed as they have a fast release cycle, while debian bullseye is already a (named) version.

Fixes #227 


11/alpine/Dockerfile
```diff
@@ -20,7 +20,8 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 
-FROM alpine:3.15.0 as jre-build
+ARG ALPINE_TAG
```
timja, line 23: I suspect you'll want to add a default value here (and in the other files for the tests), otherwise you can update the tests to pass the arg as well but generally you set defaults so that a dockerfile can be built self contained for ease of testing (it doesn't matter if the tag here doesn't get updated)

krufab: Yes, thanks for the suggestion. Locally, the tests were ok, but here clearly they failed.

docker-bake.hcl
```diff
@@ -78,8 +86,11 @@ target "alpine_jdk8" {
   }
   tags = [
```
timja, line 87: Missing passing the alpine build arg above

timja (approved):

PR #280, opened 2022-09-03
Remove jdk17-preview from one more image

timja:
missed in https://github.com/jenkinsci/docker-agent/pull/275


MarkEWaite (approved):
Thanks!

Issue #304, opened 2022-10-27
Test Harness: remove JDK17 WebStart specific test


dduportal:
Ref. https://github.com/jenkinsci/docker-agent/pull/299#discussion_r1003360580 for the principal branch.

PR #317, opened 2022-11-12
chore(linux tests) remove deprecated jdk17-only check

dduportal:
Fix https://github.com/jenkinsci/docker-agent/issues/304

### DIFF
--- a/tests/tests.bats
+++ b/tests/tests.bats
@@ -75,7 +75,7 @@ ARCH=${ARCH:-x86_64}
 @test "[${SUT_IMAGE}] use build args correctly" {
   cd "${BATS_TEST_DIRNAME}"/.. || false
 
-  local TEST_VERSION="3.36"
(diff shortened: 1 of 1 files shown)


timja (changes requested):
This is removing the check rather than the  if guard that skipped this part of the test on Java 17

dduportal:
> This is removing the check rather than the if guard that skipped this part of the test on Java 17

My bad, I misunderstood the initial intent. Should be fixed now.

timja (approved):

dduportal:
Ah ah nice one: the `use build args correctly` test, which uses another version of the remoting than the specified one, fails the associated check: it uses a remoting version pre-dating the fix https://github.com/jenkinsci/remoting/pull/532 :D 

Gotta update the "old remoting" version being checked in this PR.

oleg-nenashev (approved):

PR #399, opened 2023-04-13
fix(tests) install (non-interactively) Nuget provider as a pre-requisite for Pester installation

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3508

This PR ensures, on Windows only, during the test phase, that the `NuGet` package provider is installed as a pre-requisite for `Pester` framework.

The "subtle" part is that it requires the flag `-ForceBootstrap` because it's run non interactively in the CI environment.

The Infra team will add the `Nuget` installation in the next iteration of the ci.jenkins.io agent Vm template, but it won't kill to have this instruction also part of the `build.ps1`.

If we decide to merge this change, I'll take care of cherry-picking to other Docker images

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -120,6 +120,7 @@ if($target -eq "test") {
 
     # Only fail the run afterwards in case of any test failures
     $testFailed = $false
+    Get-PackageProvider NuGet -ForceBootstrap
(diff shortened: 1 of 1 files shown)


dduportal:
Closing in favor of https://github.com/jenkins-infra/packer-images/pull/599

PR #409, opened 2023-05-06
chore: factorize JDK specification to docker bakefile (linux) / compose file (windows)

dduportal:
[Work in progress]

Blocked by https://github.com/jenkinsci/docker-agent/pull/414

This pull request aims at factorizing the image definitions (`Dockerfile`s) to avoid duplicating code between JDK versions.

Until now, changing an element in any of the images which have both JDK11 and JDK17 definitions were subject to a risk of shift if the 2nd image is forgotten.

**How does it work?**

- The directories `11/` and `17/`, representing the major JDK version, are removed
- There is only 1 directory per operating system at the root of the repository
  - Please note that `bullseye` is renamed to `debian` during this change for the following reasons:
    - Stay homogeneous with  https://github.com/jenkinsci/docker and https://github.com/jenkinsci/docker-inbound-agent
    - The distribution codename is already specified in the `Dockerfile` as part of the `FROM` instruction and is subject to change in the futures
    - Changing the directory name breaks the GitHub permalinks: I'm proposing to tie both changes as this PR also breaks the GitHub permalinks
- The JDK version is passed through build arguments
  - The `Dockerfile`s do have a default version defined to ensure it's buildable "as it" (e.g. without Docker BuildX and its bake file). The proposed convention is to set it to the "most recent LTS JDK of the images"..
- Archlinux only exists for JDK11. It's updated to follow the same rules, and adding a JDK17 version should be another PR (and will be needed AND a good test for this refactoring).

TODO:

- [x] Factorize Linux images
- [x] Factorize Windows images
- [ ] Fix Updatecli manifests for tracking JDK versions
- [ ] Open an issue or a PR for Archlinux JDK17

### Submitter checklist


timja:
any way to see if anyone actually uses Archlinux?

dduportal:
> any way to see if anyone actually uses Archlinux?

The DockerHub download stats would be useful there

dduportal:
Rebased on top of #414 : let's see if it works as expected

MarkEWaite (approved):
Looks very good to me.

dduportal:
Please note that the following fixup changes were needed after this PR:
- https://github.com/jenkinsci/docker-agent/pull/417
- https://github.com/jenkinsci/docker-agent/pull/420
- https://github.com/jenkinsci/docker-agent/commit/e4c81a853a34112a61ba3f5aebd088b858039626

Issue #415, opened 2023-05-13
Alpine container is labeled 3.16 but delivers Alpine 3.17.3


MarkEWaite:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
Jenkins: 2.387.3
OS: Linux - 5.19.0-40-generic
Java: 11.0.19 - Eclipse Adoptium (OpenJDK 64-Bit Server VM)
---
PrioritySorter:4.1.1-rc506.e9d3571a_3387
adoptopenjdk:1.5
analysis-model-api:11.2.0
ant:487.vd79d090d4ea_e
antisamy-markup-formatter:159.v25b_c67cd35fb_
apache-httpcomponents-client-4-api:4.5.14-150.v7a_b_9d17134a_5
authentication-tokens:1.53.v1c90fd9191a_b_
authorize-project:1.5.1
badge:1.9.1
basic-branch-build-strategies:71.vc1421f89888e
bootstrap5-api:5.2.2-3
bouncycastle-api:2.27
branch-api:2.1071.v1a_188a_562481
build-blocker-plugin:1.7.8
build-name-setter:2.2.0
build-timeout:1.30
buildresult-trigger:0.18
caffeine-api:3.1.6-115.vb_8b_b_328e59d8
checks-api:2.0.0
cloudbees-bitbucket-branch-source:800.va_b_b_9a_a_5035c1
cloudbees-folder:6.815.v0dd5a_cb_40e0e
cloudbees-jenkins-advisor:336.v4d00382fe22c
clover:4.13.0
code-coverage-api:4.5.0
command-launcher:100.v2f6722292ee8
commons-lang3-api:3.12.0-36.vd97de6465d5b_
commons-text-api:1.10.0-36.vc008c8fcda_7b_
conditional-buildstep:1.4.2
config-file-provider:938.ve2b_8a_591c596
configuration-as-code:1625.v27444588cc3d
configurationslicing:539.va_85da_25f3074
copyartifact:698.v393f578eb_ddc
credentials:1254.vb_96f366e7b_a_d
credentials-binding:604.vb_64480b_c56ca_
dark-theme:315.va_22e7d692ea_a
data-tables-api:1.13.3-3
description-setter:1.10
design-library:237.v18384c2f41db
display-url-api:2.3.7
docker-commons:419.v8e3cd84ef49c
docker-workflow:563.vd5d2e5c4007f
durable-task:506.v1b_3e14b_6f5da_
[...]

PR #416, opened 2023-05-13
Label Alpine container with correct version number

MarkEWaite:
## Label Alpine container with correct version

The Alpine container image uses the Alpine version that is provided by the Elclipse Temurin Alpine container image.  That container image was upgraded from 3.16 to 3.17 months ago.  We didn't detect that version upgrade. We don't have a test that checks the Alpine version matches the tag we are creating on Dockerhub.

This is a short term solution, since it doesn't add any check to verify the Alpine version included in the Eclipse Temurin Alpine container matches the version we use to create the Dockerhub tag.

Since Alpine 3.18.0 has released, we can expect that the next Eclipse Temurin update (11.0.20, 17.0.8) in about 3 months will deliver Alpine 3.18.0 instead of 3.17.3.

Fixes https://github.com/jenkinsci/docker-agent/issues/415

### Testing done

Confirmed that Alpine 3.17 is included in the Temurin package and that the label written to Dockerhub is extracted from the variable in the bake file.

### Submitter checklist

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -62,7 +62,7 @@ variable "ON_TAG" {
 }
 
 variable "ALPINE_FULL_TAG" {
-  default = "3.16.2"
+  default = "3.17.3"
 }
 
 variable "ALPINE_SHORT_TAG" {


dduportal (approved):
Lgtm !

Issue #418, opened 2023-05-17
Code simplification: removal of default build `ARG` values when factorized


dduportal:
While working on #417 and #409  , I wondered if we could not simplify the "spread" of build arg values to factorized locations (1 for all Linux and one for all Windows).

The goal is to avoid any leftover when updating a JDK, Git or any important component.

 @team-docker-packaging I remember the constraint of having a value for build-arg in the `Dockerfile` because of the DockerHub automated builds (major) and allowing users to build a standalone `Dockerfile` without the need for additional tools (minor).

But today (May 2023):
- We do not use the DockerHub automated build since quite some time
- Both BuildX and Compose are part of any default Docker installation (or available as `docker` CLI client-side plugins easy to install for edge cases) even on Windows and Linux

Any objection for a code cleanup to remove, only when specified through a centralize file(s) , the values from `Dockerfile` ?

timja:
No objection to cleaning up the code

PR #421, opened 2023-05-17
fix(alpine) The Alpine Dockerfile does not use the `ALPINE_TAG` arg

gounthar:
@MarkEWaite detected in #415 that there was a mismatch in the naming of our Alpine Docker image, and solved it in #416.
@dduportal took some time to guide me in figuring out the problem may need a more permanent fix.

We are getting the Alpine version shipped with Eclipse Temurin. We don't take into account the `ALPINE_TAG` set in the `docker-bake.hcl` file. Damien's proposal was to do as we do with Debian, proceed with two stages.
One stage to get the Temurin build, and one to get the correct (i.e. defined in the `docker-bake`) Alpine version.
The (optional?) next step would be to use `jlink` as we do in the Debian image.

### Testing done

The only testing I have done is run locally after building the image:

```bash
docker run --rm jenkins/agent:alpine-jdk17 cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.17.3
PRETTY_NAME="Alpine Linux v3.17"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```
and 

```bash
docker run --rm jenkins/agent:alpine-jdk17 java --version
openjdk 17.0.7 2023-04-18
OpenJDK Runtime Environment Temurin-17.0.7+7 (build 17.0.7+7)
OpenJDK 64-Bit Server VM Temurin-17.0.7+7 (build 17.0.7+7, mixed mode, sharing)
```

### Submitter checklist

- [X] Link to relevant issues in GitHub or Jira
- ~~Ensure you have provided tests - that demonstrates feature works or fixes the issue~~

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -20,7 +20,13 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 ARG JAVA_VERSION=17.0.7_7
-FROM eclipse-temurin:"${JAVA_VERSION}"-jdk-alpine
+ARG ALPINE_TAG=3.17
+FROM eclipse-temurin:"${JAVA_VERSION}"-jdk-alpine AS jre-build
+
+RUN cp -r /opt/java/openjdk /javaruntime
+
+ARG ALPINE_TAG=3.17
+FROM alpine:"${ALPINE_TAG}" AS build
 
 ARG user=jenkins
 ARG group=jenkins
@@ -54,6 +60,11 @@ ARG VERSION=3107.v665000b_51092
 ADD --chown="${user}":"${group}" "https://repo.jenkins-ci.org/public/org/jenkins-ci/main/remoting/${VERSION}/remoting-${VERSION}.jar" /usr/share/jenkins/agent.jar
 RUN chmod 0644 /usr/share/jenkins/agent.jar \
   && ln -sf /usr/share/jenkins/agent.jar /usr/share/jenkins/slave.jar
+
+
+ENV JAVA_HOME=/opt/java/openjdk
+COPY --from=jre-build /javaruntime "$JAVA_HOME"
+ENV PATH="${JAVA_HOME}/bin:${PATH}"
 
 USER "${user}"
 ENV AGENT_WORKDIR="${AGENT_WORKDIR}"
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -20,7 +20,7 @@
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
 ARG JAVA_VERSION=17.0.7_7
-ARG ALPINE_TAG=3.17
+ARG ALPINE_TAG=3.17.3
 FROM eclipse-temurin:"${JAVA_VERSION}"-jdk-alpine AS jre-build
 
 RUN if [ "$TARGETPLATFORM" != 'linux/arm/v7' ]; then \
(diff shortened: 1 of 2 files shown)


dduportal:
Nice job buddy! It looks like it's working.

I'ts almost ready to merge!

I see two last steps (which we did not discuss before):

- Required for this PR to land: Could you set the `ARG ALPINE_TAG`  value to the same exact value as the variable `ALPINE_FULL_TAG` in the bake file please? (ref. https://github.com/jenkinsci/docker-agent/blob/e4c81a853a34112a61ba3f5aebd088b858039626/docker-bake.hcl#L65). The reason is to avoid any risk of configuration shift and for the sake of homogeneity (please note that #418 should solve this problem)

- Nice to have (can be in this PR or a subsequent one): track the Alpine (full) version by adding a new `updatecli` manifest.
  - You may start by duplicating https://github.com/jenkinsci/docker-agent/blob/master/updatecli/updatecli.d/jdk17.yaml which should give you an idea of the process
  - The "source" of available versions for Alpine, you can inspire yourself from this `updatecli` manifest I wrote for the Asciidoctor Docker image: https://github.com/asciidoctor/docker-asciidoctor/blob/941226d8ff0e654c3a32b4e0d0b940eddc917670/updatecli/updatecli.d/alpine-linux.yml#L16-L29 (warning: you may want to remove the `pattern: "~3.17"` to avoid locking to 3.17.x version
  - The "condition" should be set verifying that a Docker image `alpine:<found version>` is available (as it is introduced by your PR here)

alpine/Dockerfile
```diff
@@ -20,7 +20,28 @@
...
+      "$strip_java_debug_flags" \
+      --no-man-pages \
+      --no-header-files \
+      --compress=2 \
+      --output /javaruntime; \
+  else \
+    cp -r /opt/java/openjdk /javaruntime; \
+  fi
+
+ARG ALPINE_TAG=3.17.3
```
dduportal, line 43:
```suggestion
```

As the `ALPINE_TAG` build argument is only consumed by the `FROM alpine ... AS build` instruction, it's only needed once, before any `FROM instruction` as per the [Dockerfile `ARG` scope documentation](https://docs.docker.com/engine/reference/builder/#scope).

Yeah I know it's a bit weird, and this change is not blocking.

dduportal (approved):
LGTM thanks!

I've added a non blocking comment: we can merge now or add my suggestion (and fix the comment later): your call :)

gounthar:
> LGTM thanks!
> 
> I've added a non blocking comment: we can merge now or add my suggestion (and fix the comment later): your call :)

Sorry, I did not react fast enough.
I'll add it to the next PR if you don't mind.

alpine/Dockerfile
gounthar: The thing is I have an error when removing the first `ARG`.

```bash
docker buildx bake alpine_jdk17 --load
[+] Building 0.2s (3/3) FINISHED
 => [internal] load .dockerignore                                                                                                                                             0.1s
 => => transferring context: 2B                                                                                                                                               0.1s
 => [internal] load build definition from Dockerfile                                                                                                                          0.0s
 => => transferring dockerfile: 3.81kB                                                                                                                                        0.0s
 => CANCELED [internal] load metadata for docker.io/library/eclipse-temurin:17.0.7_7-jdk-alpine                                                                               0.0s
Dockerfile:43
--------------------
  41 |
  42 |     ARG ALPINE_TAG=3.17.3
  43 | >>> FROM alpine:"${ALPINE_TAG}" AS build
  44 |
  45 |     ARG user=jenkins
--------------------
ERROR: failed to solve: failed to parse stage name "alpine:": invalid reference format
```

PR #445, opened 2023-06-28
refactor(build): factorize Jenkinsfile steps

lemeurherve:
This PR refactors Jenkinsfile steps in order to facilitate the introduction of Windows LTSC 2022 and future images like in https://github.com/jenkinsci/docker-agent/pull/441.

~As it would have been more difficult to do otherwise, it includes an evolution of the changes of #442 and #444.~

~Extract from https://github.com/jenkinsci/docker-agent/pull/441#issuecomment-1612214991,~
> ~I'd suggest reviewing (and hopefully merging) https://github.com/jenkinsci/docker-agent/pull/442 then https://github.com/jenkinsci/docker-agent/pull/444 then [this PR]~

These PRs have since been merged.

### Testing done

Replay with all tests passing: https://ci.jenkins.io/job/Packaging/job/docker-agent/job/PR-445/4/

Last version: https://ci.jenkins.io/job/Packaging/job/docker-agent/job/PR-445/55/

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #452, opened 2023-07-07
chore(updatecli) fixups of JDK manifests and remove depreciation warnings

dduportal:
Minor fixups of #445 + remove updatecli's depreciation warnings  avoid empty PR due to double quotes


dduportal:
Merging as updatecli worjks as expected!

Issue #451, opened 2023-07-06
Add `nanoserver` `ltsc2019` along with `1809`


dduportal:
### What feature do you want to see added?

As per https://learn.microsoft.com/en-us/windows-server/get-started/servicing-channels-comparison and https://hub.docker.com/_/microsoft-windows-nanoserver, we should provide an LTSC2019 for `nanoserver` in replacement of `1809`.

That would ensure our users always have the latest stable content from the OS base image, ant it would help us for maintenance (ref. https://github.com/jenkinsci/docker-agent/pull/445/files#r1254304996 for instance).

----

This would impact the published tags by changing the suffix `1809` by `ltsc2019`.

It means we also want to introduce a set of (additional) new tags such as  `jdk11-nanoserver`,  `3107.v665000b_51092-16-jdk11-nanoserver`, etc. like we are already doing with the Linux image while keeping the existing ones.

Note that https://docs.docker.com/compose/compose-file/build/#tags could GREATLY help in that matter

dduportal:
Editing issue title: we want both 1809 (to keep existing images) in addition to the new ltsc2019.

lemeurherve:
Reopening as the suggestion done in #475 hasn't been applied.

Should be closed by https://github.com/jenkinsci/docker-agent/pull/488

PR #459, opened 2023-07-08
feat: add `nanoserver-ltsc2019` tags

lemeurherve:
#451 

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -86,10 +86,11 @@ $baseDockerCmd = 'docker-compose --file=build-windows.yaml'
 $baseDockerBuildCmd = '{0} build --parallel --pull' -f $baseDockerCmd
 
 Invoke-Expression "$baseDockerCmd config --services" 2>$null | ForEach-Object {
-    $image = '{0}-{1}' -f $_, $env:WINDOWS_VERSION_NAME
+    $initialImage = '{0}-{1}' -f $_, $env:WINDOWS_VERSION_NAME
     # Special case for nanoserver-1809 images
-    $image = $image.replace('nanoserver-ltsc2019', 'nanoserver-1809')
+    $image = $initialImage.replace('nanoserver-ltsc2019', 'nanoserver-1809')
     $items = $image.Split("-")
+    # Remove the 'jdk' prefix (3 first characters)
     $jdkMajorVersion = $items[0].Remove(0,3)
     $windowsType = $items[1]
     $windowsVersion = $items[2]
@@ -99,6 +100,11 @@ Invoke-Expression "$baseDockerCmd config --services" 2>$null | ForEach-Object {
     $tags = @( $image, $versionTag )
     if($jdkMajorVersion -eq "$defaultJdk") {
         $tags += $baseImage
+    }
+    # Special case for nanoserver-ltsc2019 additional tag
+    if($image.Contains('nanoserver-1809')) {
+        $additionalVersionTag = "${RemotingVersion}-${BuildNumber}-${initialImage}"
+        $tags += @( $initialImage, $additionalVersionTag )
     }
 
     $builds[$image] = @{


lemeurherve:
Unfortunately the command `docker-compose --file=build-windows.yaml build --parallel --pull` called at https://github.com/jenkinsci/docker-agent/blob/b49cad5551ba80af87260fb2968030bb107c7981/build.ps1#L118-L120 for building all images at once is problematic, it doesn't take in account the `$builds` array of images and their tags and thus doesn't have any idea of the additional `nanoserver-ltsc2019` images 😓 

To fix that without a bigger refactoring would mean iterate on every `$builds` items to call the build like https://github.com/jenkinsci/docker-agent/blob/b49cad5551ba80af87260fb2968030bb107c7981/build.ps1#L113-L116

dduportal:
> Unfortunately the command `docker-compose --file=build-windows.yaml build --parallel --pull` called at
> 
> https://github.com/jenkinsci/docker-agent/blob/b49cad5551ba80af87260fb2968030bb107c7981/build.ps1#L118-L120
> for building all images at once is problematic, it doesn't take in account the `$builds` array of images and their tags and thus doesn't have any idea of the additional `nanoserver-ltsc2019` images 😓
> 
> To fix that without a bigger refactoring would mean iterate on every `$builds` items to call the build like
> 
> https://github.com/jenkinsci/docker-agent/blob/b49cad5551ba80af87260fb2968030bb107c7981/build.ps1#L113-L116

Could we run LTSC2019 and 1809 in parallel in two 2019 agents?

lemeurherve:
> > Unfortunately the command `docker-compose --file=build-windows.yaml build --parallel --pull` called at
> > https://github.com/jenkinsci/docker-agent/blob/b49cad5551ba80af87260fb2968030bb107c7981/build.ps1#L118-L120
> > 
> > for building all images at once is problematic, it doesn't take in account the `$builds` array of images and their tags and thus doesn't have any idea of the additional `nanoserver-ltsc2019` images 😓
> > To fix that without a bigger refactoring would mean iterate on every `$builds` items to call the build like
> > https://github.com/jenkinsci/docker-agent/blob/b49cad5551ba80af87260fb2968030bb107c7981/build.ps1#L113-L116
> 
> Could we run LTSC2019 and 1809 in parallel in two 2019 agents?

This should work but then we'll need to add a `windows-1809` label on (trusted.)ci.jenkins.io agents so I can use the agent label to generate the corresponding images and tags.

dduportal:
> EDIT: will set agent label dynamically instead.

Yup that was the (not really explicit :D)  idea: `if endsWith('1809')  then set to 2019)

Jenkinsfile
```diff
@@ -11,14 +24,14 @@ pipeline {
             matrix {
                 axes {
                     axis {
-                        name 'AGENT_TYPE'
-                        values 'linux', 'windows-2019', 'windows-2022'
+                        name 'IMAGE_TYPE'
+                        values 'linux', 'nanoserver-1809', 'nanoserver-ltsc2022', 'windowsservercore-ltsc2019', 'windowsservercore-ltsc2022'
```
lemeurherve, line 28: To fix #451 there will be an additional PR containing this change:
```suggestion
                        values 'linux', 'nanoserver-1809', 'nanoserver-ltsc2019', 'nanoserver-ltsc2022', 'windowsservercore-ltsc2019', 'windowsservercore-ltsc2022'
```

Not sure if it will be needed but we could also add a `windowsservercore-1809` "proper" docker agent image with the following change:
```suggestion
                        values 'linux', 'nanoserver-1809', 'nanoserver-ltsc2022', 'windowsservercore-1809', 'windowsservercore-ltsc2019', 'windowsservercore-ltsc2022'
```

lemeurherve: Previous build containing these 2 additional sets of images: https://ci.jenkins.io/job/Packaging/job/docker-agent/job/PR-459/55/

lemeurherve: > we could also add a "proper" windowsservercore-1809 docker agent image

Rethinking about it, we should propose this new image in this PR IMO, so we can mention in the release note that if you want to keep a WSC image as close as the previous one you can use this new `windowsservercore-1809` image to avoid breaking change.

lemeurherve: `windowsservercore-1809` image added.

build-windows.yaml
```diff
@@ -1,37 +1,23 @@
 services:
-  jdk11-nanoserver:
-    image: jdk11-nanoserver-${NANOSERVER_VERSION_NAME}
+  jdk11:
+    image: jdk11-${WINDOWS_FLAVOR}-${WINDOWS_VERSION_TAG}
     build:
-      context: ./windows/nanoserver/
+      context: ./
+      dockerfile: ./windows/${WINDOWS_FLAVOR}//Dockerfile
```
dduportal, line 6:
```suggestion
      dockerfile: ./windows/${WINDOWS_FLAVOR}/Dockerfile
```

windows/windowsservercore/Dockerfile
```diff
@@ -23,10 +23,11 @@
 #  THE SOFTWARE.
 
 ARG JAVA_VERSION=17.0.7_7
-ARG WINDOWS_VERSION_TAG=1809
-FROM eclipse-temurin:"${JAVA_VERSION}"-jdk-windowsservercore-"${WINDOWS_VERSION_TAG}" AS jdk-core
+ARG WINDOWS_VERSION_TAG=ltsc2019
+ARG WINDOWS_VERSION_FALLBACK_TAG=1809
+FROM eclipse-temurin:"${JAVA_VERSION}"-jdk-windowsservercore-"${WINDOWS_VERSION_FALLBACK_TAG}" AS jdk-core
```
dduportal, line 28:
```suggestion
FROM eclipse-temurin:"${JAVA_VERSION}"-jdk-windowsservercore-"${WINDOWS_VERSION_FALLBACK_TAG}" AS jdk-core
```

nitpicking on naming: `WINDOWS_VERSION_FALLBACK_TAG` feels hard to get. 

WDYT about `TOOLS_WINDOWS_VERSION` instead?

- prefix `TOOLS` to make it clear for what it is used (temurin and pwsh)
- no more `TAG` suffix because it is not a tag (neither Docker image tag nor git tag)

dduportal (approved):
That is a nice work! Approving the PR as it looks good to me.

I've added a nitpicking about variable naming but it should not block this PR.

I'm waiting for 2 days before merging, to let time to other maintainers if they want to weight in, then I'll merge and we'll start publishing.

build-windows.yaml
```diff
@@ -1,37 +1,23 @@
 services:
-  jdk11-nanoserver:
```
dduportal, line 2: nit: I can't remember why we "simplified" so much the compose file. As much as I want to factorize the Dockerfile (which has duplicated logic), I wonder if using the docker-compose to describe (and pass) all kinds of definition we need, like the bakefile, wouldn't help to clarify (and avoid tedious "if then else" in the powershell) ?

build-windows.yaml
lemeurherve: Nice catch!

build-windows.yaml
lemeurherve: This is related to https://github.com/jenkinsci/docker-agent/pull/459#issuecomment-1656034950

There is no "else" in the PowerShell script, what are the tedious parts you're alluding to?

dduportal: As far as I remember, https://github.com/jenkinsci/docker-agent/pull/459#issuecomment-1656034950 was before you worked into splitting the image build in different lines (in different agents) so I considered it outdated.

> There is no "else" in the PowerShell script, what are the tedious parts you're alluding to?

https://github.com/jenkinsci/docker-agent/pull/459/files#diff-3258bc4162690adead72c70b672093e68c92e86a4628148f0015d61738f21b5aR75-R77

While reviewing the PR, I realized that we store part of the definitions in the compose YAML manifest and the other part in code like this one. I have not a solution for it yet (and it is not blocking the PR) but I believe that we should be able to define the logic ltsc2019 (tools at 1809, base at ltsc2019) / 1809 (all in 1809) / ltsc2022 (all in ltsc2022) in one (or multiple) compose files to avoid splitting intelligence between "definition" and "programmatic".

I can totally take a try at this once we have started to deliver the image if ok for you?

dduportal: (don't get me wrong, your work is really good here, I'm nitpicking here!)

lemeurherve:
FTR, some updatecli manifests are now failing, will look into it later.

image

Jenkinsfile
lemeurherve: `nanoserver-ltsc2019` addition in https://github.com/jenkinsci/docker-agent/pull/488

PR #498, opened 2023-09-13
feat(platforms): Adds ppc64le and s390x archs for JDK21.

gounthar:
After discussing with @ksalerno99 yesterday during the [Platform SIG meeting](https://www.jenkins.io/sigs/platform/), I considered that adding support for a few platforms for JDK21 would be beneficial.

I then propose to add `"linux/ppc64le"`, `"linux/armv/v7"`  and `"linux/s390x"` as new target platforms for our JDK21 preview images.

### Testing done

`docker buildx bake --file docker-bake.hcl debian_jdk21`

### Submitter checklist
- ~~Link to relevant pull requests, esp. upstream and downstream changes~~

### DIFF
--- a/debian/21/Dockerfile
+++ b/debian/21/Dockerfile
@@ -1,4 +1,4 @@
-ARG BOOKWORM_TAG=20230725
+ARG BOOKWORM_TAG=20230904
 FROM debian:bookworm-"${BOOKWORM_TAG}"-slim as jre-build
 ARG JAVA_VERSION
 ARG TARGETPLATFORM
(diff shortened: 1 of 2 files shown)


debian/21/Dockerfile
```diff
@@ -13,21 +13,26 @@ RUN set -x; apt-get update \
...
-  --compress=zip-6 \
-  --output /javaruntime
+RUN if test "${TARGETPLATFORM}" != 'linux/arm/v7'; then \
+      jlink \
+        --add-modules ALL-MODULE-PATH \
+        --no-man-pages \
+        --compress=zip-6 \
+        --output /javaruntime; \
+    # It is acceptable to have a larger image in arm/v7 (arm 32 bits) environment.
+    # Because jlink fails with the error "jmods: Value too large for defined data type" error.
```
timja, line 30: did you test and see if this is still needed?

gounthar: Yes, I tested that this morning, and it's still the case.
Thanks.

timja (approved):

debian/21/Dockerfile
ksalerno99: Hi, this error you were having with jlink is caused by running a 32-bit arm qemu emulator on a 64-bit host. When jlink tries to open a large (>2GB) file mmap fails.

https://bugs.launchpad.net/qemu/+bug/1805913

gounthar: Thank you Kenneth. 🙏 
It looks like this bug will never be fixed. 🤷

PR #503, opened 2023-09-13
fix(windows) bump `git-lfs` to 3.4.0 and fix its installation

lemeurherve:
This PR fixes the `git: 'lfs' is not a git command.` error noticed in https://github.com/jenkinsci/docker-agent/pull/502#issuecomment-1717648730 when using the version 3.4.0 of git-lfs by adding git to the `PATH` earlier then using git-lfs.exe installer instead of `git.exe lfs install`.

It also adds a test ensuring git-lfs (and thus git) is correctly installed.

Supersedes and closes #502

### Testing done

https://ci.jenkins.io/job/Packaging/job/docker-agent/job/PR-503/11/

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #506, opened 2023-09-15
chore(updatecli) correct regexp error to fix git-lfs manifest

dduportal:
This PR is an `updatecli` fixup of #503 for the `git-lfs` manifest


dduportal:
Merging as updatecli check is 👍

lemeurherve:
Thanks!

PR #520, opened 2023-10-11
breaking: set Java 17 as default

lemeurherve:
This PR sets Java 17 as the default JDK and short tags like what has been done in the official Docker image: https://github.com/jenkinsci/docker/pull/1724

Changes:
- Set JDK17 as default in Linux and Windows builds for short tags
- Add `latest-bookworm` tag (to be coherent with the other Linux flavors)
- ~~Add archlinux-17 images (related: #422)~~ removed in https://github.com/jenkinsci/docker-agent/pull/520/commits/7fc8b833084800037957267c4cbd8c8a379d50b4, can be reverted or added in another PR if needed.
- Update README with JDK17 images as default short tags (closes #489)

### Testing done

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #518, opened 2023-10-11
feat(Windows): add Java 21

lemeurherve:
This PR adds jdk21 variants for Windows images.

### Testing done

https://ci.jenkins.io/job/Packaging/job/docker-agent/job/PR-518/3/

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #521, opened 2023-10-12
feat(Linux): switch supported architectures to GA Temurin jdk21 images

lemeurherve:
This PR switch supported architectures (x86 & arm64) to the GA jdk21 Temurin Linux images.

It keeps jdk21-preview images for unsupported architectures (ppc64le, arm32 and s390x).

It also include a fix for the preview (EA) release download URL.

Follow-up of #518 

~~Depending on #520 decision about archlinux, I could add an archlinux-jdk21 image too.~~ removed from that PR.

### Testing done

https://ci.jenkins.io/job/Packaging/job/docker-agent/job/PR-521/32/

```[tasklist]
### Submitter checklist
```


gounthar (commented):
Maybe we should wait, then, until they become available.

lemeurherve:
> Maybe we should wait, then, until they become available.

I can rework my PR to keep linux-arm32 and linux-ppc64le as jdk21-**preview** images, WDYT?

timja (changes requested):
~use a multi-arch image to retrieve the jdk please instead of fetching it via wget like we do in all other images~

I was looking at an old diff

docker-bake.hcl
```diff
@@ -211,8 +211,25 @@ target "debian_jdk17" {
...
+    equal(ON_TAG, "true") ? "${REGISTRY}/${JENKINS_REPO}:${REMOTING_VERSION}-${BUILD_NUMBER}-jdk21" : "",
+    "${REGISTRY}/${JENKINS_REPO}:bookworm-jdk21",
+    "${REGISTRY}/${JENKINS_REPO}:jdk21",
+    "${REGISTRY}/${JENKINS_REPO}:latest-bookworm-jdk21",
+    "${REGISTRY}/${JENKINS_REPO}:latest-jdk21",
+  ]
+  platforms = ["linux/amd64", "linux/arm64"]
+}
+
+target "debian_jdk21_preview" {
```
timja, line 232: we should remove this and re-add when it's available, I don't think there's much point in shipping these preview tags in the meantime, it's confusing.

docker-bake.hcl
```diff
@@ -211,8 +211,25 @@ target "debian_jdk17" {
...
+    DEBIAN_RELEASE = DEBIAN_RELEASE
+  }
+  tags = [
+    equal(ON_TAG, "true") ? "${REGISTRY}/${JENKINS_REPO}:${REMOTING_VERSION}-${BUILD_NUMBER}-jdk21" : "",
+    "${REGISTRY}/${JENKINS_REPO}:bookworm-jdk21",
+    "${REGISTRY}/${JENKINS_REPO}:jdk21",
+    "${REGISTRY}/${JENKINS_REPO}:latest-bookworm-jdk21",
+    "${REGISTRY}/${JENKINS_REPO}:latest-jdk21",
+  ]
+  platforms = ["linux/amd64", "linux/arm64"]
```
timja, line 229: I would probably just wait a couple of days they probably won't take long to produce the other platforms?

lemeurherve:
> > Maybe we should wait, then, until they become available.
> 
> I can rework my PR to keep linux-arm32 and linux-ppc64le as jdk21-**preview** images, WDYT?
> 
> EDIT: done.

There aren't any binaries for these architectures in recent releases, and I'm not sure there will ever be.

To keep preview images for these specific architecture without blocking the major ones, I propose to use the (fixed) last release where they're available in https://github.com/jenkinsci/docker-agent/blob/f8e2ca6cf61fca89c90896d839c1dd0547aaff3a/debian/21/Dockerfile#L8-L19

WDYT?

lemeurherve:
We can also wait a few days.

timja (approved):

gounthar:
Thank you.
Here is the current status found `[here](https://github.com/adoptium/temurin/issues/8#issue-1933244087)`.
image

lemeurherve:
> Thank you. Here is the current status found [here](https://github.com/adoptium/temurin/issues/8#issue-1933244087). 
Thanks for the link, quite useful!

> Could we wait for `arm32`, `ppc64le` [...] to go?

Yes, I'll keep this PR as draft until `arm32` & `ppc64le` images variants are ready.

>  and `x64 Windows`

Unrelated to this PR concerning Linux images.

We're already publishing amd64 Windows images (see https://hub.docker.com/r/jenkins/agent/tags?page=1&name=21), which AFAIU are the same architecture.

> I can't find the right link at the moment, but there is an explanation somewhere in a Temurin repo where they describe the different classes of platforms they support. I guess `s390x` is in the last group.

I've removed `s390x` target architecture for jdk21 as it won't be supported in commit https://github.com/jenkinsci/docker-agent/pull/521/commits/2d4b31908c3af3fb86018a623f61b32996b240d6

lemeurherve:
https://github.com/adoptium/temurin/issues/8 has been closed, and the October release status doesn't look good for the architectures we were waiting for: https://github.com/adoptium/temurin/issues/6

image

With these new elements at hand, are you OK if I restore a similar workaround as the one deployed in https://github.com/jenkinsci/docker/pull/1738 to keep the preview tags/images like in https://github.com/jenkinsci/docker-agent/pull/521/commits/2eae83657f4016b0d9b95ea7fefbbd6c89065318?

MarkEWaite:
I think that the workaround is a very reasonable compromise.

dduportal:
+1 with the workaround! Will be double justified as there is a ne EA release for `s390x` for instance, as pointed by @gounthar : https://github.com/adoptium/temurin21-binaries/releases/tag/jdk-21.0.1%2B12-ea-beta

gounthar:
Yes, `ppc64le`, `s390x` and `armv7` are available for the JDK 21.0.1 early access.

lemeurherve:
I'll prepare the workaround, then as follow-up of this PR I'll prepare another one for updatecli manifests managing both type of versions like https://github.com/jenkinsci/docker/pull/1741

lemeurherve:
Included back the workaround, and also incorporated the same kind of preview (EA) release download URL fix as https://github.com/jenkinsci/docker/pull/1747

PR ready for review, to be squashed.

dduportal (approved):

lemeurherve:
> I'll prepare the workaround, then as follow-up of this PR I'll prepare another one for updatecli manifests managing both type of versions like [jenkinsci/docker#1741](https://github.com/jenkinsci/docker/pull/1741)

Updatecli manifests update addressed in https://github.com/jenkinsci/docker-agent/pull/536

PR #551, opened 2023-10-27
cleanup(Windows): include `TOOLS_WINDOWS_VERSION` in image type

lemeurherve:
This PR simplifies a bit `build.ps1` script by including the `TOOLS_WINDOWS_VERSION` in image type.

### Testing done

```[tasklist]
### Submitter checklist
```


dduportal (commented):
I'm not approving neither refusing this PR as I don't see the problem it solves (or I don't understand or don't agree) but it is not a blocker: if other maintainers agrees to merge this PR I don't want to block the process.

----

- What is the problem solved by this change? It's not immediately clear to me or I don't understand the "simplification" stated by the PR body message. 
- The current implementation is hard to understand when checking the `Jenkinsfile` unless you are a maintainer with a lot of implementation details in your mind: `nanoserver-1809` is easy to get, but `nanoserver-1809-1809` is not: is it a typo? What is the intent.
  - Eventually `nanoserver-1809-tools-1809`? Or at least a comment explaining the convention and mechanisms at stake here?
- Personal opinion: The choice of the OSes to select for the set of "docker stage images used to retrieve binary-compatible tools" for a given image should be kept encapsulated inside the build tools (wether Dockerfile, Docker Bake in the future or the `builds.ps1` script today): moving this intelligence out of the build tool to the pipeline creates coupling making the maintenance really hard: You must use Jenkins with a Jenkinsfile to reproduce a build locally unless you are a maintainer with all the knowledge, creating a big step for newcomers or any help


=> I would rather keep the current code as it is an exception for only a small subset. The `if` is currently well commented from my point of view and give a clear explanation in the proper location in term of "digging in this repository". 
Besides, if we can switch to docker-compose or docker bake, making this exception in declarative should be easy.

lemeurherve:
Closing this PR as after sleeping on it I'm not satisfied either.

Issue #760, opened 2024-03-06
updatecli is not working for updates to JDKs


tmousawNF:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
  Paste the output here
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

Agent is running  Linux

### Reproduction steps

I have an issue with an image that is based off jenkins/inbound-agent:latest-jdk21 because an older version of JDK 21 is used. I believe https://bugs.openjdk.org/browse/JDK-8320072 is causing my tests to fail. This bug is fixed in version 21.0.2_8. However, the last bump in JDK version was via https://github.com/jenkinsci/docker-agent/pull/560 which updated it to 21.0.1_12. I believe changes to the conditions in updatecli for the JDK to open a pull request is causing them not to occur. Further, images for JDK 21 are being published only for Ubuntu 22.04 (jammy) whereas the conditions are looking for Ubuntu 20.04 (focal).

### Expected Results

updatecli automatically runs once a week and opens pull requests for any updated JDKs for JDK 11, 17, and 21.

### Actual Results

updatecli has not been opening pull requests since its last one on October 30th, 2023.

### Are you interested in contributing a fix?

I do believe I have a fork with changes that will fix this.

gounthar:
In conditions, I see:

```
checkTemurinDebianDockerImages
------------------------------
ERROR: GET https://index.docker.io/v2/library/eclipse-temurin/manifests/21.0.2_13-jdk-focal: MANIFEST_UNKNOWN: manifest unknown; unknown tag=21.0.2_13-jdk-focal
```

tmousawNF:
Exactly. It's because of the OS. I think I have the fix and am just about to put up a pull request.

gounthar:
As you please, I'm also working on a PR. ;-)

gounthar:
Ok, #761 and #762 aren't the same, but I guess maintainers will take #761 because it addresses more issues. 👍 
Well done! 🙏

MarkEWaite:
Resolved by the merge.  Thanks very much!

PR #761, opened 2024-03-06
Issue #760 - updatecli changes to enable JDK updates when new JDK versions are published

tmousawNF:
For [issue 760](https://github.com/jenkinsci/docker-agent/issues/760).

Updates made to the conditions of the JDK specific updatecli configs back in October have caused them to never pass and, therefore, no JDK updates have been made since October.

Most of the changes simply relate to the fact that the OS was not listed in the architectures for the conditions. While testing out the changes, I also found the following changes that needed to be made:
- JDK 11's current version is jdk-11.0.22+7.1 which is an update to jdk-11.0.22+7. The pattern matching matched this version and caused failures. This is because the pattern needed to escape the '+' symbol such that it matched jdk-11.0.22+7, but not jdk-11.0.22+7.1.
- I made the same update to JDK 17's updatecli config as I made to JDK11.
- JDK 21 was looking for Ubuntu 20.04 (focal) based docker images. JDK 21 only appears to publish Ubuntu 22.04 (jammy) based docker images.

### Testing done

I tested locally by doing a dry run using updatecli:
```
# updatecli diff --config ./updatecli/updatecli.d --values ./updatecli/values.github-action.yaml
```

This allowed me to view what updates would be made.

I was not entirely certain how (or if) an automated test could be written for this piece.

```[tasklist]
### Submitter checklist
```


tmousawNF:
The build failed on a 503 (Unavailable). I'm curious if this was a network blip. I found a typo in one of the existing configs, so will be updating the PR which should trigger another build. If 503 happens again, then I'll take a closer look.

gounthar:
> The build failed on a 503 (Unavailable). I'm curious if this was a network blip. I found a typo in one of the existing configs, so will be updating the PR which should trigger another build. If 503 happens again, then I'll take a closer look.

I also got a 503 with different changes on Windows. 🤷 
I guess this may be a network glitch.

tmousawNF:
What are next steps here? Looks like I need someone with write access to merge this, but I'm unsure how to find out who has write access.

gounthar:
It's team-docker-packaging.
They will review when time permits.

MarkEWaite:
> * JDK 11's current version is jdk-11.0.22+7.1 which is an update to jdk-11.0.22+7. The pattern matching matched this version and caused failures. This is because the pattern needed to escape the '+' symbol such that it matched jdk-11.0.22+7, but not jdk-11.0.22+7.1.

The JDK 11 version that you mentioned is specific to aarch64 for macOS.  The other JDK 11 versions (amd64, etc.) are still at 11.0.22+7.  Will your changes handle that case?

tmousawNF:
> > * JDK 11's current version is jdk-11.0.22+7.1 which is an update to jdk-11.0.22+7. The pattern matching matched this version and caused failures. This is because the pattern needed to escape the '+' symbol such that it matched jdk-11.0.22+7, but not jdk-11.0.22+7.1.
> 
> The JDK 11 version that you mentioned is specific to aarch64 for macOS. The other JDK 11 versions (amd64, etc.) are still at 11.0.22+7. Will your changes handle that case?

If you look at the [JDK 11 releases page](https://github.com/adoptium/temurin11-binaries/releases), the aarch64 and macOS versions published under 11.0.22+7.1 have the same URL as the ones published under 11.0.22+7. My assumption is that the "dot" versions are hot fixes that get published under the same URL for some critical bug. Therefore, finding the latest non "dot" version should still work since the same URL is used. If I am misunderstanding how this works, then please let me know.

tmousawNF:
I just pushed some changes that I think will allow the handling of hot fix releases. I applied the changes to jdk11.yaml, jdk17.yaml, and jdk21.yaml and all seem to pass locally.

updatecli/updatecli.d/jdk11.yaml
```diff
@@ -37,55 +37,73 @@ conditions:
   checkTemurinAlpineDockerImage:
     kind: dockerimage
     name: Check if the container image "eclipse-temurin:<lastVersion>-jdk-alpine" is available
-    disablesourceinput: true
+    transformers:
+      - findsubmatch:
+          pattern: "(.*\\_\\d*)(\\.\\d*)?$"
+          captureindex: 1
+      - addsuffix: "-jdk-alpine"
```
dduportal, line 44: It looks like this transformation works as expected:

```
checkTemurinAlpineDockerImage
-----------------------------
✔ docker image eclipse-temurin:11.0.22_7-jdk-alpine found
```
as per the last builds.

However the targets do not benefit from this:

```
⚠ - change detected:
	* key "$.services.jdk11.build.args.JAVA_VERSION" should be updated from "\"11.0.20.1_1\"" to "11.0.22_7.1", in file "build-windows.yaml"
```

same build.


=> We might want to factorize all the transformers directly into the source (at least the pattern selection) so no need to maintain all of these across conditions and targets, WDYT?

=> Which raise the question: with only the fix on the architectures (which I would use with the long form everywhere so it is homogeneous and easier to get), shouldn't the version`11.0.22_7` be picked by the source without the combination of (changed regex + removing last digit with transformer to handle the specific aarch64 case)?

dduportal: > It looks like this transformation works as expected:
> 
> ```
> checkTemurinAlpineDockerImage
> -----------------------------
> ✔ docker image eclipse-temurin:11.0.22_7-jdk-alpine found
> ```
> 
> as per the last builds.
> 
> However the targets do not benefit from this:
> 
> ```
> ⚠ - change detected:
> 	* key "$.services.jdk11.build.args.JAVA_VERSION" should be updated from "\"11.0.20.1_1\"" to "11.0.22_7.1", in file "build-windows.yaml"
> ```
> 
> same build.
> 
> => We might want to factorize all the transformers directly into the source (at least the pattern selection) so no need to maintain all of these across conditions and targets, WDYT?
> 
> => Which raise the question: with only the fix on the architectures (which I would use with the long form everywhere so it is homogeneous and easier to get), shouldn't the version`11.0.22_7` be picked by the source without the combination of (changed regex + removing last digit with transformer to handle the specific aarch64 case)?

I see in #762 that the current source also picks `jdk-11.0.22+7.1`.

```
✔ GitHub release version "jdk-11.0.22+7.1" found matching pattern "^jdk-11.(\\d*).(\\d*).(\\d*)(.(\\d*))+(\\d*)$" of kind "regex"
```

This exotic schemes looks like a problem!

@tmousawNF @gounthar @MarkEWaite @timja 

WDYT about keeping the current regexp for source, but add the transformer proposed in this PR to remove any additional number after the `+` ?

tmousawNF: > It looks like this transformation works as expected:
> 
> ```
> checkTemurinAlpineDockerImage
> -----------------------------
> ✔ docker image eclipse-temurin:11.0.22_7-jdk-alpine found
> ```
> 
> as per the last builds.
> 
> However the targets do not benefit from this:
> 
> ```
> ⚠ - change detected:
> 	* key "$.services.jdk11.build.args.JAVA_VERSION" should be updated from "\"11.0.20.1_1\"" to "11.0.22_7.1", in file "build-windows.yaml"
> ```
> 
> same build.
This is exactly what I intended. When a hot fix is published, the docker image will be tagged without the hot fix portion (i.e. `11.0.22_7` instead of `11.0.22_7.1` in this particular case) such that the previously published docker image is replaced with the newer docker image. However, the actual "JAVA_VERSION" we're building the docker images for will be based on the hot fix, so I think it is correct to update the "JAVA VERSION" to include the hot fix portion. Otherwise, if we had built the image for version `11.0.22_7`, how would we trigger the docker image to build again to pick up the hot fix? It's possible I'm misunderstanding how this works, but my assumption is that the update of the "JAVA_VERSION" triggers the docker image to re-build. Please let me know if I have a misunderstanding here.

> 
> => We might want to factorize all the transformers directly into the source (at least the pattern selection) so no need to maintain all of these across conditions and targets, WDYT?
> 
Yes. This is a very practical suggestion. I'll work on an update to accomplish this.

> => Which raise the question: with only the fix on the architectures (which I would use with the long form everywhere so it is homogeneous and easier to get), shouldn't the version `11.0.22_7` be picked by the source without the combination of (changed regex + removing last digit with transformer to handle the specific aarch64 case)?
If I understand your question, it is true that version `11.0.22_7` will be picked up without the change when the `11.0.22_7` version is published. However, as I stated in my previous response, when `11.0.22_7.1` is published, without the changed regex how is the hot fix built into a docker image? Again, I could be misunderstanding what triggers the build of the docker image and, if so, please let me know.

tmousawNF: Actually, now that I try to build with these suggested changes, I see that this is an issue because the `JAVA_VERSION` is pulling the docker image. How do we ensure that docker images get built when hot fixes are published?

olblak: Hey,
Small suggestion would it make sense to also pin using the docker image digest something like `eclipse-temurin:11.0.22_7-jdk-alpine@sha256xxxxx`

dduportal: > Actually, now that I try to build with these suggested changes, I see that this is an issue because the JAVA_VERSION is pulling the docker image. How do we ensure that docker images get built when hot fixes are published?

When `updatecli` detects the change, it opens a PR with the proposed changed. It only automates what a human would have done manually.

=> then the PR mechanism is responsible to verify that the "change" can be built.
=> if the PR is approved and merged, then the "change" is on the staging list until a new release is created (usually once a week).


That is why we expect `updatecli` to change the `JAVA_VERSION` variables (https://github.com/jenkinsci/docker-agent/blob/79d99e606eb5c30838aa8e77ba2c6c5e4fd8fd5f/docker-bake.hcl#L111-L125 and https://github.com/jenkinsci/docker-agent/blob/79d99e606eb5c30838aa8e77ba2c6c5e4fd8fd5f/build-windows.yaml#L9 for windows as example) which are passed to the docker build process as arguments.

As today we expect the Docker build argument `JAVA_VERSION` to specify the image to use, **if** a hotfix is valid we expect to catch it for all version and use it for all.

Does it answer your question?

dduportal: > Hey, Small suggestion would it make sense to also pin using the docker image digest something like `eclipse-temurin:11.0.22_7-jdk-alpine@sha256xxxxx`

Hi @olblak , what would be the problem to solve by using the digest exactly? Not sure to see the reason here

olblak: Sorry maybe I misunderstood the initial problem here. I thought that challenge was to detect "hostfix" update in version like `11.0.22_7.1`

Feel free to ignore my suggestion then

tmousawNF: > As today we expect the Docker build argument `JAVA_VERSION` to specify the image to use, **if** a hotfix is valid we expect to catch it for all version and use it for all.
> 
> Does it answer your question?

What I would like to understand is how the image for a hot fix will be incorporated into new docker images. For this particular example, the version `11.0.22+7.1` will publish a new version of the docker image with tag `11.0.22_7` (same tag as the non-hot fix version). My question is what would spur a new build to occur if no other changes were made to this repository? Are weekly releases made even if there are no changes?

gounthar: There is no docker image for the `11.0.22+7.1` (as it is macOS-only), so there would not be a new docker image from this repo for this version. 🤷

dduportal: > Sorry maybe I misunderstood the initial problem here. I thought that challenge was to detect "hostfix" update in version like `11.0.22_7.1`
> 
> Feel free to ignore my suggestion then

No problem @olblak I believe the discussion is mandatory here to ensure the best contribution from @tmousawNF and it also benefits the community! Thanks for weighing in! I believe it is another use case for https://github.com/updatecli/updatecli/issues/1733 (which might interest you)

tmousawNF: With respect to @gounthar's comment:

> There is no docker image for the 11.0.22+7.1 (as it is macOS-only), so there would not be a new docker image from this repo for this version. 🤷

I found an example of where there was a "hotfix" that included new docker images (see https://github.com/adoptium/temurin17-binaries/releases/tag/jdk-17.0.9%2B9.1).

I'm going to suggest that I will remove the detection of the "hotfix" versions and only detect the major versions. In the meantime, we are no worse off than before since the "hotfix" causes issues in the existing config file (just try to run the `jdk11.yaml` config today).

I have also yet to factorize all the transformers. I will work on that now.

dduportal: I remember this one, true! 

There was no associated Docker image: https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=17.0.9_9.1 (compared to https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=17.0.9_9).

I wonder about their packaging policy in these cases: do they override the existing images with the hotfix content?

- If yes then we should not have to worry about catching the hotfix: next release of the Jenkins images will automatically embed the "hotifx" (the tip from @olblak could apply in these cases: if we retrieve the SHA checksum we should be able to have updatecli to help by opening a PR with the same Docker image tag BUT a different checksum)
- If no, then we should not care about.

Anyway, thanks for your work: it alrerady fixes a LOT of cases and improves the situation so let's roll once you're finished with the refacto!

updatecli/values.temurin.yaml
```diff
@@ -0,0 +1,2 @@
+temurin:
+  version_pattern: "^jdk-[11|17|21].(\\d*).(\\d*).(\\d*)(.(\\d*))\\+(\\d*)?$"
```
dduportal, line 2:
```suggestion
  version_pattern: "^jdk-[11|17|21].(\\d*).(\\d*).(\\d*)(.(\\d*))\\+(\\d*)?$"

```

(nit) [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html) says we must have an empty line at the end of a file.

dduportal (approved):
Nice work, thanks!

The result seen in the CI checks looks really good:

```
setJDK11VersionDockerBake
-------------------------

**Dry Run enabled**

⚠ - changes detected:
	path "variable.JAVA11_VERSION.default" updated from "11.0.21_9" to "11.0.22_7" in file "docker-bake.hcl"

setJDK11VersionWindows
----------------------

**Dry Run enabled**

⚠ - change detected:
	* key "$.services.jdk11.build.args.JAVA_VERSION" should be updated from "\"11.0.20.1_1\"" to "11.0.22_7", in file "build-windows.yaml"


setJDK17VersionWindows
----------------------

**Dry Run enabled**

⚠ - change detected:
	* key "$.services.jdk17.build.args.JAVA_VERSION" should be updated from "\"17.0.8.1_1\"" to "17.0.10_7", in file "build-windows.yaml"

setJDK17VersionDebian
---------------------

**Dry Run enabled**

⚠ The line #23, matched by the keyword "ARG" and the matcher "JAVA_VERSION", is changed from "ARG JAVA_VERSION=17.0.9_9" to "ARG JAVA_VERSION=17.0.10_7".
⚠ - changed lines [23] of file "/tmp/updatecli/github/jenkinsci/docker-agent/debian/Dockerfile"

setJDK17VersionAlpine
---------------------

**Dry Run enabled**

⚠ The line #22, matched by the keyword "ARG" and the matcher "JAVA_VERSION", is changed from "ARG JAVA_VERSION=17.0.8.1_1" to "ARG JAVA_VERSION=17.0.10_7".
⚠ - changed lines [22] of file "/tmp/updatecli/github/jenkinsci/docker-agent/alpine/Dockerfile"

setJDK17VersionDockerBake
-------------------------

**Dry Run enabled**

⚠ - changes detected:
	path "variable.JAVA17_VERSION.default" updated from "17.0.9_9" to "17.0.10_7" in file "docker-bake.hcl"


setJDK21PreviewVersionDockerBake
--------------------------------

**Dry Run enabled**

⚠ - changes detected:
	path "variable.JAVA21_PREVIEW_VERSION.default" updated from "21.0.1+12" to "21.0.3+6" in file "docker-bake.hcl"


setJDK21Version
---------------

**Dry Run enabled**

⚠ - changes detected:
	path "variable.JAVA21_VERSION.default" updated from "21.0.1_12" to "21.0.2_13" in file "docker-bake.hcl"
```


I've added only a nitpicking comment (non blocking).


I would want to have another maintainer review before merging

updatecli/values.temurin.yaml
tmousawNF: I had to merge with `master` anyway, so I updated this file.

dduportal (approved):

MarkEWaite (approved):

PR #779, opened 2024-03-21
chore(updatecli) Tracks Ubuntu and Debian LTS releases

gounthar:
The aim of this draft PR is to update our `Dockerfiles`, `docker-bake.hcl`, and `README_AGENT.md` to the latest LTS versions of Ubuntu and Debian whenever they are released (approximately every two years).

I understand it may seem like a significant effort for a change that occurs only every two years. However, this should prompt us to maybe reconsider our use of a combination of Debian and Ubuntu.

Currently, we base our Debian image on a Temurin image (Ubuntu-based) to obtain the JDK binaries. Then, we utilize `jlink` to create a customized version of the JDK. Finally, we start with a Debian image and integrate this customized JDK into it.
Is there a guarantee that there is perfect compatibility between the `libc` used in the Ubuntu image and the one used in the Debian image? Not to my knowledge.
While it undeniably works, we should maybe question why and for how long.
Does Temurin confirm that the `libc` in their Ubuntu image is compatible with the one provided with the Debian image?
No.
While Ubuntu is based on Debian (Ubuntu builds on the Debian architecture and infrastructure), I believe we may be overly confident in this regard.

Temurin provides packages through standard installers for various Linux distributions, should we go this way?
Temurin also provides [binaries](https://adoptium.net/installation/archives/#_linux) in the `tgz` format for various CPU architectures running Linux. There is also no assurance that these binaries would function with Debian, but they do.
Is this because they are statically linked, or is it due to Temurin building them on updated Debian machines and because we also use the current Debian release?
I am uncertain, but this approach appears more reasonable to me.
Additionally, in the event of a CVE, Temurin releases its updated binaries much quicker than the Docker images.
Therefore, perhaps we should consider eliminating the Ubuntu step and directly utilizing the binaries.

### Testing done

```
updatecli diff --debug --config ./updatecli/updatecli.d/debian-new-LTS.yaml --values ./updatecli/values.github-action.yaml --values ./updatecli/values.temurin.yaml
updatecli diff --debug --config ./updatecli/updatecli.d/debian-codename.yaml --values ./updatecli/values.github-action.yaml --values ./updatecli/values.temurin.yaml
updatecli diff --debug --config ./updatecli/updatecli.d/ubuntu.yaml --values ./updatecli/values.github-action.yaml --values ./updatecli/values.temurin.yaml
```

### Submitter checklist
- ~~Link to relevant pull requests, esp. upstream and downstream changes~~


gounthar:
The build failure happens on Windows, so it shouldn't be linked to this change.

dduportal:
Hi @gounthar , what is the status of this PR? Is it still needed?

gounthar:
Hi @dduportal,

We can close, this needs further investigation.
Thanks!

PR #786, opened 2024-04-15
feat(Linux): switch from temurin base images to temurin installer

lemeurherve:
This PR switches from temurin base images to the temurin installer in Linux images.

Benefits:
- Less layers to pull
- Faster availability as we don't need to wait for every kind of images (OS/architecture) to be released
- Simpler updatecli manifests

Notes:
- Reuse the jdk-download scripts introduced in https://github.com/jenkinsci/docker/pull/1847
- The JDK21 preview build failed with the jdk download scripts, kept as is since it's already a bit different than the other Linux Dockerfiles.
- Extracted in #799:
  - ~Removed the TARGETPLATFORM check in Alpine images as we're not publishing any arm/v7 image of this kind~
  - ~Integrate jlink improvements from https://github.com/jenkinsci/docker/pull/1857~

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4029#issuecomment-2045073813

### Testing done
Local builds + CI

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -55,7 +55,6 @@ RUN case "$(jlink --version 2>&1)" in \
       --add-modules ALL-MODULE-PATH \
       --no-man-pages \
       --no-header-files \
-      --compress=2 \
       --output /javaruntime; \
   else \
     cp -r "/opt/jdk-${JAVA_VERSION}" /javaruntime; \
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -56,9 +56,6 @@ RUN case "$(jlink --version 2>&1)" in \
       --no-man-pages \
       --no-header-files \
       --output /javaruntime; \
-  else \
-    cp -r "/opt/jdk-${JAVA_VERSION}" /javaruntime; \
-  fi
 
 ## Agent image target
 FROM alpine:"${ALPINE_TAG}" AS agent
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -55,7 +55,7 @@ RUN case "$(jlink --version 2>&1)" in \
       --add-modules ALL-MODULE-PATH \
       --no-man-pages \
       --no-header-files \
-      --output /javaruntime; \
+      --output /javaruntime
 
 ## Agent image target
 FROM alpine:"${ALPINE_TAG}" AS agent
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -19,26 +19,43 @@
 #  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
-ARG JAVA_VERSION=17.0.10_7
 ARG ALPINE_TAG=3.19.1
-FROM eclipse-temurin:"${JAVA_VERSION}"-jdk-alpine AS jre-build
+FROM alpine:"${ALPINE_TAG}" AS jre-build
 
-RUN if [ "$TARGETPLATFORM" != 'linux/arm/v7' ]; then \
-    case "$(jlink --version 2>&1)" in \
+SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
+
+# This Build ARG is populated by Docker
+# Ref. https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
+ARG TARGETPLATFORM
+
+COPY jdk-download-url.sh /usr/bin/local/jdk-download-url.sh
+COPY jdk-download.sh /usr/bin/local/jdk-download.sh
+
+ARG JAVA_VERSION=17.0.10_7
+# hadolint ignore=DL3018
+RUN apk add --no-cache \
+    ca-certificates \
+    jq \
+    curl \
+  && /usr/bin/local/jdk-download.sh alpine
+
+ENV PATH="/opt/jdk-${JAVA_VERSION}/bin:${PATH}"
+
+RUN case "$(jlink --version 2>&1)" in \
       # jlink version 11 has less features than JDK17+
-      "11."*) strip_java_debug_flags="--strip-debug" ;; \
-      *) strip_java_debug_flags="--strip-java-debug-attributes" ;; \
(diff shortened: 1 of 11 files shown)


alpine/Dockerfile
```diff
@@ -19,9 +19,35 @@
 #  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 #  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 #  THE SOFTWARE.
-ARG JAVA_VERSION=17.0.10_7
 ARG ALPINE_TAG=3.19.1
-FROM eclipse-temurin:"${JAVA_VERSION}"-jdk-alpine AS jre-build
+FROM alpine:"${ALPINE_TAG}" AS jre-build
+ARG JAVA_VERSION
```
dduportal, line 24: Is there a reason to move the `ARG JAVA_VERSION` inside the scope of the first "stage"?

In theory the second stage does not have any code to consume its value (and `ARG` value are not persisted in the final image) so it should be OK. But given how confusing the `ARG` scope in Dockerfile (ref. https://docs.docker.com/reference/dockerfile/#scope) I propose to keep the `ARG JAVA_VERSION` instruction on top so any stage can consume it

alpine/Dockerfile
```diff
@@ -19,9 +19,35 @@
...
+    ca-certificates \
+    jq \
+    wget \
+  && DASHED_JAVA_VERSION=$(echo "$JAVA_VERSION" | sed -e 's/+/_/') \
+  && JAVA_MAJOR_VERSION=$(echo "$JAVA_VERSION" | cut -d'+' -f1 | cut -d'.' -f1) \
+  && ENCODED_JAVA_VERSION=$(echo "$JAVA_VERSION" | jq "@uri" -jRr) \
+  && CONVERTED_ARCH=$(arch | sed -e 's/x86_64/x64/' -e 's/armv7l/arm/') \
+  && wget --quiet https://github.com/adoptium/temurin"${JAVA_MAJOR_VERSION}"-binaries/releases/download/jdk-"${ENCODED_JAVA_VERSION}"/OpenJDK"${JAVA_MAJOR_VERSION}"U-jdk_"${CONVERTED_ARCH}"_alpine-linux_hotspot_"${DASHED_JAVA_VERSION}".tar.gz -O /tmp/jdk.tar.gz \
+  && tar -xzf /tmp/jdk.tar.gz -C /opt/ \
+  && rm -f /tmp/jdk.tar.gz
```
dduportal, line 48: See proposal in https://github.com/jenkinsci/docker/pull/1847/files#r1567144856 (as way to ensure reusability without going imemdiatly into the whole "common parent image" pattern)

hervelemeur: Note that this code was already a part of this repo (Idem about the ARG location), which I slightly optimised and reduced.

https://github.com/jenkinsci/docker-agent/blob/e2560a929884a1f1cfe3c64001b1b2932d8e7057/debian/preview/Dockerfile#L3-L29

I've got a better proposal: retrieve the binary directly from api.adoptium.net, offering a much better querying interface which doesn't need all these java version variations, WDYT?

alpine/Dockerfile
hervelemeur: > Is there a reason to move the ARG JAVA_VERSION inside the scope of the first "stage"?

Being iso with the existing code:
https://github.com/jenkinsci/docker/pull/1847/files#r1567144856

alpine/Dockerfile
dduportal: if it works, that would simplify a lot!

gounthar: My first tests with the [API](https://api.adoptium.net/q/swagger-ui/#/Installer/getInstallerByVersion) aren't working really well.
The default example to get an installer gives me a `404`:
```
curl -X 'GET' \
  'https://api.adoptium.net/v3/installer/version/jdk-11.0.6%2B10/linux/x64/jdk/hotspot/normal/eclipse?project=jdk' \
  -H 'accept: */*'
```

dduportal (commented):
Given the size of the PR (which will make it a nightmare to review), how do you feel about splitting Linux/Windows into 2 distinct PRs so we can focus on 1 paltform then on the other?

lemeurherve:
Note that splitting this PR in two parts will result in updatecli failing until both will be merged: I'm using the original version containing `+`, not converted to `_`.

Also, that PR is (IMO) not that much complicated:
- All Linux image changes is more or less the same
- The Windows images needed a new parameter passed at test time, the JAVA_VERSION

Happy to do an interactive review if needed, can also split the PR anyway, your call 🙂

alpine/Dockerfile
dduportal: With the new "Temurin API call", it is still a LOT of code (the search and replace) repeated across the dockerfile. I maintain my proposal to factorize into a single shell script

dduportal:
#787 has been merged: you can proceed with this one (gotta spend some time on it to ensure it is delivered soon).

alpine/Dockerfile
```diff
@@ -37,7 +55,10 @@ RUN case "$(jlink --version 2>&1)" in \
       --add-modules ALL-MODULE-PATH \
       --no-man-pages \
       --no-header-files \
-      --output /javaruntime
+      --output /javaruntime; \
+  else \
+    cp -r "/opt/jdk-${JAVA_VERSION}" /javaruntime; \
+  fi
```
dduportal, line 61: Fails with the followi!ng error:

```
[2024-05-07T06:05:25.856Z] #79 [agent_alpine_jdk17 jre-build 5/5] RUN case "$(jlink --version 2>&1)" in       "11."*) set -- "--strip-debug" "--compress=2" ;;       "17."*) set -- "--strip-java-debug-attributes" "--compress=2" ;;       "21."*) set -- "--strip-java-debug-attributes" "--compress=zip-6" ;;       *) echo "ERROR: unmanaged jlink version pattern" && exit 1 ;;     esac;     jlink       "$1"       "$2"       --add-modules ALL-MODULE-PATH       --no-man-pages       --no-header-files       --output /javaruntime;   else     cp -r "/opt/jdk-17.0.10_7" /javaruntime;   fi
[2024-05-07T06:05:25.856Z] #79 0.285 /bin/ash: syntax error: unexpected "else"
```

=> there are no `if` directive

lemeurherve: Doh, wrong upstream merge... Fix incoming.

dduportal (approved):

PR #787, opened 2024-04-17
feat(Windows): switch from temurin base images to temurin installer

lemeurherve:
This PR switches from temurin base images to the temurin installer for the Windows images.

Benefits:
- Less layers to pull
- Faster availability as we don't need to wait for every kind of images (OS/architecture) to be released

Notes:
- Apart from the use of the installer instead of a temurin base image is the change to the build process to pass the JAVA_VERSION from the docker compose file to the tests.
- Using the happy path of api.adoptium.net, retrieving the MSI installer from jdk version only (retrieving it from the GitHub releases requires the use of several variation of the jdk version)
- Using a Windows Server Core `jdk-core` base image for the nanoserver image: we can't use a nanoserver one as we need to retrieve some files from a Windows Server Core anyway
  https://github.com/jenkinsci/docker-agent/blob/e2560a929884a1f1cfe3c64001b1b2932d8e7057/windows/nanoserver/Dockerfile#L40-L41
  The alternative would be to retrieve the jdk from an existing Windows Server Core agent image and thus having to wait for it to be available first as trade-off.
  FTR, that's how Adoptium is making their own nanoserver image: https://github.com/adoptium/containers/blob/df93f85c97eff616bf43de2c6ff7661d4174f833/21/jdk/windows/nanoserver-1809/Dockerfile#L34
- Deliberately not including jlink in this pull request for Windows images as there is currently nothing in the repo capable of testing it properly

Split of:
- #786 

The updatecli manifests will be updated at once after switching off Windows and Linux images.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4029#issuecomment-2045073813
- https://github.com/jenkinsci/docker-agent/pull/786#pullrequestreview-2005125722

### Testing done
Local builds + CI

```[tasklist]
### Submitter checklist
```


dduportal (approved):
🚀

PR #802, opened 2024-05-07
Add `-noReconnectAfter` option support

Vlatombe:
* [ ] Downstream of https://github.com/jenkinsci/remoting/pull/738

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/jenkins-agent
+++ b/jenkins-agent
@@ -37,6 +37,7 @@
 # * JENKINS_INSTANCE_IDENTITY: The base64 encoded InstanceIdentity byte array of the Jenkins controller. When this is set,
 #                              the agent skips connecting to an HTTP(S) port for connection info.
 # * JENKINS_PROTOCOLS:         Specify the remoting protocols to attempt when instanceIdentity is provided.
+# * JENKINS_NO_RECONNECT_AFTER: Do not attempt to reconnect to the Jenkins controller after the given timeout.
 
 if [ $# -eq 1 ] && [ "${1#-}" = "$1" ] ; then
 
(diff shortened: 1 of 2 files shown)


jglick:
Do we really need this? I feel like we should deprecate the env vars in this launcher script except for one catch-all `REMOTING_OPTS` var with whatever CLI options you like; then https://github.com/jenkinsci/kubernetes-plugin/blob/f10083a42e708a6bff990c6ccb306ce618b8bccd/src/main/java/org/csanchez/jenkins/plugins/kubernetes/PodTemplateBuilder.java#L415-L449 could (eventually, once adopted) be simplified to just construct its list of CLI options directly without needing to make further changes to this image beyond routine JAR bumps.

jglick:
Yes, and for compatibility we need to continue to support the currently defined variables, I am just suggesting that going forward we deprecate them in favor of a generic launcher options variable. Would also allow us to deprecate and eventually remove messy stuff like https://github.com/jenkinsci/docker-agent/blob/611dc9df849eee9334a3351a6a520e2106116dc3/jenkins-agent#L60-L61 https://github.com/jenkinsci/docker-agent/blob/611dc9df849eee9334a3351a6a520e2106116dc3/jenkins-agent#L131-L132

You can already pass any arguments you like to the entrypoint (`"$@"`) but for purposes of some cloud impls like `PodTemplateBuilder` it is easier to set environment variables.

Vlatombe:
> any arguments you like to the entrypoint

Only on linux, the windows launcher script is missing this option.

Filed https://github.com/jenkinsci/docker-agent/pull/809 as a draft.

Issue #777, opened 2024-03-18
DockerHub Description automated process is broken since 2 months (at least?)


dduportal:
Each time there is a release, the automated process used to push the updated README to DockerHub is failing:

```
Acquiring token
Sending PATCH request
Error: Forbidden
```

It's been at least 2 months with this failure:

![Capture d’écran 2024-03-18 à 18 41 53](https://github.com/jenkinsci/docker-agent/assets/1522731/f9e36046-78d8-45b9-84d8-1712f0226232)

I'm not sure why, and how to fix it yet (haven't looked at it) but opening issue for awareness

cc @lemeurherve @MarkEWaite @NotMyFault @gounthar @timja

itayporezky:
I reached here to submit a issue to update the description and saw this,
I believe the dockerhub description is about ~7 months out of date... seeing as the description has mentions of bullseye and no mentions of bookworm.

lemeurherve:
I don't have access to the jenkins docker hub account to check if the GitHub repository secrets are valid.

Looking at the code of the GHA used to update description and short description on hub.docker.com at https://github.com/peter-evans/dockerhub-description/blob/main/src/dockerhub-helper.ts, I'm wondering if we shouldn't do these 2 API calls ourselves in a pipeline library function (for example) where Docker hub username and password are already used.

That would avoid setting up and keep up to date GitHub secrets on this repository.

WDYT?

MarkEWaite:
> That would avoid setting up and keep up to date GitHub secrets on this repository, and reuse this new function in other docker repositories in @jenkinsci & @jenkins-infra orgs.
> 
> WDYT?

That seems very reasonable to me.  If the credentials are already used in the Pipeline job and if there is a relatively easy API call that will do the work, then let's call the API directly from the Pipeline.

timja:
The reason it was done like this is because the delete permission is required and delete wasn’t wanted on the main credential

dduportal:
What @timja said. Separating these credentials is key to safety (and avoid mistakes)

timja:
I've fixed this by setting the right permissions on the agent image for robots (admin).

inbound-agent already had the right permissions


-----

Yes the permissions aren't ideal but we've gone through this at length previously and it's a docker hub limitation.

timja:
> I reached here to submit a issue to update the description and saw this, I believe the dockerhub description is about ~7 months out of date... seeing as the description has mentions of bullseye and no mentions of bookworm.

[...]

Issue #422, opened 2023-05-18
Decide the future of the Archlinux image: either add "Archlinux JDK17" or remove "Archlinux JDK11"


dduportal:
### What feature do you want to see added?

There is currently an Archlinux JDK11 image introduced in https://github.com/jenkinsci/docker-agent/pull/171.

We've already wondered if we (Jenkins community) should keep maintaining it or remove it:

- There are no co-maintainer specified in https://github.com/jenkinsci/docker-agent/blob/master/.github/CODEOWNERS for this image so it falls back to "general maintainers".
- There is a JDK11 version but no JDK17 of this image
- There are no Archlinux version for jenkinsci/docker-inbound-agent - https://github.com/jenkinsci/docker-inbound-agent/pull/301
- There no Archlinux version for jenkinsci/docker-ssh-agent
- There no Archlinux version for jenkinsci/docker (controller)

As discussed in https://github.com/jenkinsci/docker-agent/pull/409#issuecomment-1537207386 , let's start by checking numbers: how much pull of the archlinux image do we have and which part of the usage does it represent?
If it is really low, unless some steps up to help maintaining, we should consider deprecating this image.

Ping @fossdd @slide as you were the author/maintainer of the PR introducing Alpine Linux: do you feel like maintaining it or deprecating it?

gounthar:
Do you have any numbers to share @dduportal ?
I'd love to keep the ArchLinux image (and move it to JDK17), but only if it is used by the community.

lemeurherve:
Archlinux will be removed as soon as https://github.com/jenkinsci/docker-agent/pull/890 is delivered.

MarkEWaite:
> Archlinux will be removed as soon as [#890](https://github.com/jenkinsci/docker-agent/pull/890) is delivered.

Platform SIG meeting discussed further and agreed that merging next week is good.  Arch Linux container images will no longer be updated beginning next week.

PR #803, opened 2024-05-07
chore(doc): Reformat readme tags

itayporezky:
(Draft, see open questions below)
### Background:
After using the docker agents for much time, i found i was actually using less optimal tags then the ones i randomly discovered in the dockerhub tag search

### Description
1. This is a reformatting of the readme to hopefully make the tags that are available to be used more clear. 
2. Also added 1 more tag in the bake file to align with the same amount of tags created on alpine and same format.
3. Open questions:
- which i thought about too late: Is there an actual need for so many JDK flavors?
- Deprecate archlinux? released tags seem old

(I suggest to review using some Markdown previewer)
Related #777 

### Testing done
N/A

```[tasklist]
### Submitter checklist
- [X] Link to relevant issues in GitHub or Jira
```

### DIFF
--- a/README_agent.md
+++ b/README_agent.md
@@ -126,7 +126,7 @@ The image has several supported configurations, which can be accessed via the fo
 The file [docker-bake.hcl](https://github.com/jenkinsci/docker-agent/blob/master/docker-bake.hcl) defines all the configuration for Linux images and their associated tags.
 
 There are also versioned tags in DockerHub, and they are recommended for production use.
-See the full list at [https://hub.docker.com/r/jenkins/agent/tags)](https://hub.docker.com/r/jenkins/agent/tags)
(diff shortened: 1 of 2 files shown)


timja (approved):
I've pushed some general improvements, ideally tagged builds would be promoted in some way, they are quite far down and probably not as clear as they should be but I think this is better than before so approving.

timja:
> Deprecate archlinux? released tags seem old

See https://github.com/jenkinsci/docker-agent/issues/422

itayporezky:
Looks good for now :)

timja:
FWIW in case you are interested, inbound agent is in an even worse state with no mentioned tags
https://github.com/jenkinsci/docker-agent/blob/master/README_inbound-agent.md

https://hub.docker.com/r/jenkins/inbound-agent

Issue #798, opened 2024-05-06
JDK 21 preview updatecli manifest is failing


lemeurherve:
### Jenkins and plugins versions report

N/A

### What Operating System are you using (both controller, and any agents involved in the problem)?

N/A

### Reproduction steps

Observe updatecli actions logs in this repository.

### Expected Results

No error in logs.

### Actual Results

Error in logs:

> SOURCES
> =======
> 
> getLatestJDK21PreviewVersion
> ----------------------------
> Searching for version matching pattern ".*-ea-.*"
> ✔ Git tag "jdk-21.0.4+1-ea-beta" found matching pattern ".*-ea-.*" of kind "regex"
> 
> 
> CONDITIONS:
> ===========
> 
> checkIfReleaseIsAvailable
> -------------------------
> The shell 🐚 command "/bin/sh /var/folders/ly/77mv7l9968g068w9_fwzwggw0000gp/T/updatecli/bin/d3ffbc3fbbe15c1cb14dacbe20856423b65a42ac0add7e5e19d1b29ad3fe138d.sh" exited on error (exit code 1) with the following output:
> ----
> ERROR: the following URL is NOT available: https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21.0.4%2B1-ea-beta/OpenJDK21U-jdk_x64_linux_hotspot_21.0.4_1-ea.tar.gz.

lemeurherve:
Not sure why it's failing on main, on #796 but not on #786, not on #799 and not locally 🤔

gounthar:
`updatecli` version maybe?

dduportal:
The error is clearly related to the shell script:

```
<...> exited on error (exit code 1) with the following output:
```

=> worth executing it in debug over there

gounthar:
The script is failing, but the URL is valid. 🤔

lemeurherve:
> The script is failing, but the URL is valid. 🤔

It was returning a 404 error yesterday 🤔

Maybe a GitHub intermittent issue that would explain why it succeeded sometimes?

lemeurherve:
FTR, the URL from #772 isn't the same cf https://ci.jenkins.io/job/Packaging/job/docker-agent/job/PR-772/16/pipeline-console/?selected-node=102

```diff
[...]

PR #806, opened 2024-05-14
chore: allow running updatecli with `--debug`

lemeurherve:
This PR adds an input to the workflow_dispatch of the updatecli GitHub Action so we can use it to run updatecli with the `--debug` parameter.

Ref:
- https://github.com/jenkinsci/docker-agent/issues/798#issuecomment-2097656708

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -2,6 +2,12 @@ name: updatecli
 on:
   # Allow to be run manually
   workflow_dispatch:
+    inputs:
+      # Optional debug input
+      debug:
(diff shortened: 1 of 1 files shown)


dduportal (changes requested):
I'm not completely sure (to confirm with @olbak) but doing so would risk exposing sensitive values such as the GitHub token so at first sight I'm not voting for this change.

Additionally, I don't understand why this would be helpful for https://github.com/jenkinsci/docker-agent/issues/798#issuecomment-2097656708 as the issue was regarding the shell script itself, not by updatecli.

PR #809, opened 2024-05-15
Introduce `REMOTING_OPTS` to pass arbitrary options to remoting on startup via an environment variable

Vlatombe:
https://github.com/jenkinsci/docker-agent/pull/802#issuecomment-2098889201

This deprecates existing environment variables allowing to set options to `agent.jar`. Going forward, this will ease addition or removal of options to `agent.jar` without impacting these launcher scripts.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/jenkins-agent
+++ b/jenkins-agent
@@ -37,6 +37,7 @@
 # * JENKINS_INSTANCE_IDENTITY: The base64 encoded InstanceIdentity byte array of the Jenkins controller. When this is set,
 #                              the agent skips connecting to an HTTP(S) port for connection info.
 # * JENKINS_PROTOCOLS:         Specify the remoting protocols to attempt when instanceIdentity is provided.
+# * REMOTING_OPTS:             Generic way to pass additional options to pass to the remoting library
 
 if [ $# -eq 1 ] && [ "${1#-}" = "$1" ] ; then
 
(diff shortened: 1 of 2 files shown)


jenkins-agent
```diff
@@ -37,6 +37,7 @@
 # * JENKINS_INSTANCE_IDENTITY: The base64 encoded InstanceIdentity byte array of the Jenkins controller. When this is set,
 #                              the agent skips connecting to an HTTP(S) port for connection info.
 # * JENKINS_PROTOCOLS:         Specify the remoting protocols to attempt when instanceIdentity is provided.
+# * REMOTING_OPTS:             Generic way to pass additional options to pass to the remoting library
```
jglick, line 40:
```suggestion
# * REMOTING_OPTS:             Generic way to pass additional CLI options to agent.jar (see -help)
```

jglick (approved):
> Should be decided what to do when both systems (command-line options, specific environment variables) are used and fail-fast accordingly.

I do not think we need to make it unnecessarily complicated; there is already too much code that tries to be clever this way, but `Launcher` itself should detect inappropriate option combinations and produce appropriate error messages.

I _would_ suggest moving the new option to the top of the description, and noting that it supersedes all prior options except for `JAVA_BIN` and `JAVA_OPTIONS` and that use of those is now deprecated.

jglick (approved):

jglick:
@lemeurherve ?

dduportal (approved):

lemeurherve (approved):

PR #857, opened 2024-08-14
Bugfix: Fix regression where environment variables set to USER would be lowercased to user.

aidanleuck:
https://github.com/jenkinsci/docker-agent/issues/858 (this will be moved at some point)

Fixes issue with environment variables named USER being set to user due to it being merged with the variable set in the Dockerfile.

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira (Waiting for issue to be moved)
- [ N/A ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -105,7 +105,7 @@ RUN mkdir -p /home/"${user}"/.jenkins && mkdir -p "${AGENT_WORKDIR}"
 VOLUME /home/"${user}"/.jenkins
 VOLUME "${AGENT_WORKDIR}"
 WORKDIR /home/"${user}"
-ENV user=${user}
+ENV USER=${user}
 LABEL \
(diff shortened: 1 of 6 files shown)


dduportal:
Closes https://github.com/jenkinsci/docker-inbound-agents/issues/44

dduportal (approved):
Many thanks! that looks good!

dduportal (changes requested):
@aidanleuck you also have to update the test harness: checks are failing due to this

aidanleuck:
@dduportal Thanks I fixed those tests. It appears it is still failing trying to download JDK, looking at past builds it appears this might be an intermittency as opposed to changes in my PR. Do we just need to keep running it again?

dduportal (approved):
LGTM!

Issue #883, opened 2024-09-25
Add an updatecli manifest to track UBI9 docker image version


lemeurherveCB:
### What feature do you want to see added?

#866 introduced UBI9 agents, refined in #880 but there is no updatecli manifest to keep `UBI9_TAG` up to date in the Dockerfile and in the docker-bake.hcl files.

I got a first version that I removed in https://github.com/jenkinsci/docker-agent/pull/880/commits/ea742576977655f45756b82bd50d10569397f862

Without any additional spec the last tag returned is `sha256-14f1cd4b565c93b7ae400db1759db485f14a9154d45f25c9558cd2a9e753d2ca.att`(wrong one):
```yaml
sources:
  latestVersion:
    kind: dockerimage
    name: "Get the latest UBI9 Linux version"
    spec:
      image: "registry.access.redhat.com/ubi9/ubi"
```

Using `versionfilter: semver` + `trimsuffix: "-source"` also returns wrong version:
```yaml
sources:
  latestVersion:
    kind: dockerimage
    name: "Get the latest UBI9 Linux version"
    spec:
      image: "registry.access.redhat.com/ubi9/ubi"
      versionfilter:
        kind: semver
    transformers:
      - trimsuffix: "-source"
```
> The line #1, matched by the keyword "ARG" and the matcher "UBI9_TAG", is changed from "ARG UBI9_TAG="9.4-1214.1726694543" to "ARG UBI9_TAG=9.4".

Same with a regexp pattern, wrong version returned.
```yaml
sources:
  latestVersion:
    kind: dockerimage
    name: "Get the latest UBI9 Linux version"
    spec:
      image: "registry.access.redhat.com/ubi9/ubi"
      versionfilter:
        kind: regexp
        pattern: \d+\.\d+-\d+\.\d+
    transformers:
      - trimsuffix: "-source"
```
> The line #1, matched by the keyword "ARG" and the matcher "UBI9_TAG", is changed from "ARG UBI9_TAG="9.4-1214.1726694543" to "ARG UBI9_TAG=9.4-947.1717074712".

MarkEWaite:
Fixed by

* #885

PR #885, opened 2024-09-26
feat(updatecli): Track UBI9 version

gounthar:
After [@lemeurherve's work](https://github.com/jenkinsci/docker-agent/pull/880) on the `updatecli` manifest to keep UBI9 updated in the docker bake file and Dockerfile, it became clear something wasn't working as expected when trying to find the latest version of the UBI9 container with the default `updatecli` behavior.  

I then took Hervé's manifest and transitioned it into using a shell script to find the "right" version.  
I discovered a [Swagger description](https://catalog.redhat.com/api/containers/v1/ui/#/Repositories/graphql.images.get_images_by_repo) that led me to the correct API endpoint.  
This endpoint allowed me to query the `ubi9` containers, sorted by the last update date and filtered by the `latest` tag.  
As a result, I obtained the version `9.4-1214.1726694543`.  

The script works as is but would need some additional refinement to return a list of versions rather than just the most recent one.

This could help partially solve #883.

### Testing Done

I updated my local configuration to target my fork instead of upstream.  
To test the changes, I intentionally kept my master branch out-of-date, and `updatecli` successfully created [this PR](https://github.com/gounthar/docker-agent/pull/1) for me.

The command used for testing:  
`updatecli diff --debug --config ./updatecli/updatecli.d/rhel-ubi9.yaml --values ./updatecli/values.github-action.yaml 2>&1`

### Submitter checklist

### DIFF
--- a/updatecli/scripts/ubi9-latest-tag.sh
+++ b/updatecli/scripts/ubi9-latest-tag.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 
-# The Swagger API endpoints for the Red Hat Container Catalog API are documented at: https://catalog.redhat.com/api/containers/v1/ui/#/Repositories/graphql.images.get_images_by_repo
+# This script fetches the latest tag from the Red Hat Container Catalog API for UBI9 images.
+# It ensures that `jq` and `curl` are installed, fetches the tags, and processes them to find the unique tag.
+
+# The Swagger API endpoints for the Red Hat Container Catalog API are documented at:
+# https://catalog.redhat.com/api/containers/v1/ui/#/Repositories/graphql.images.get_images_by_repo
+
 # Correct URL of the Red Hat Container Catalog API for UBI9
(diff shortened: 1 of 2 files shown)


timja (commented):
this seems a bit over complicated, does renovate or dependabot do this correctly and simpler?

gounthar:
I think dependabot handles the version retrieval pretty well, but I'm not sure it can update the docker bake file.

timja:
> I think dependabot handles the version retrieval pretty well, but I'm not sure it can update the docker bake file.

renovate would be able to with the regex method pretty easily then in the bake file

dduportal:
Please note that this "issue" should be temporarily and updatecli should be fixed

lemeurherve:
This PR added a renamed version of ./updatecli/scripts/ubi9-tags.sh (introduced in https://github.com/jenkinsci/docker-agent/pull/901) named ./updatecli/scripts/ubi9-tags.sh, but didn't remove the original, now unused.

PR #960, opened 2025-03-26
chore(Jenkinsfile) add support for retries with non Spot VMs on ci.jenkins.io

dduportal:
This PR changes the pipeline to support the new infrastructure agent labels pattern on ci.jenkins.io (ref. https://github.com/jenkins-infra/helpdesk/issues/4598 and https://github.com/jenkins-infra/pipeline-library/pull/917).

- Switch to scripted syntax to support the `retry()` with counter and agent failure detection (see the note below)
- Ensure that `trusted.ci.jenkins.io` Linux VM agent labels are not used on ci.jenkins.io which result in failed build due to CPU mismatch on ci.jenkins.io (since the `linux` label is too generic)

(edit) NOTE: The label `docker-highmem` is now using Spot instances (ref. https://github.com/jenkins-infra/helpdesk/issues/4598). In the current state (even with #961).
If the spot agent is reclaimed, the build fails on ci.jenkins.io.

This PR is about retrying the build, on ci.jenkins.io, when the underlying agent is failing (spot reclaimed, deletion by the infra team, cloud failure, etc.).

It requires the agent label to change during the pipeline execution which is not possible using the declarative syntax, as it is parses the (static) `agent` block directive at the beginning.
Also, even if `retry` can be specified to the `options` with the proposed argument, it does not provide a dynamic counter.
=> these 2 reasons explains why a shift to scripted is needed.

----

Given the discussion (which I did not anticipated) towards scripted vs. declarative, it means we have to decide if we want the pipeline to:

- Retry if agent failure is detected, but with the same label (e.g. again with spot). But we can keep declarative.
- Or go toward my proposal and switch to scripted.
- Or eventually find a solution (and dismiss my arguments) to fallback to `nonspot` as part of the pipeline.
----

In the future, we might want to provide an `infra.withNode()` in the jenkins-infra/pipeline-library to provide an higher order of abstraction (encapsulating retries, spot selections, etc.)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,7 @@ parallelStages = [failFast: false]
     'windowsservercore-ltsc2019',
     'windowsservercore-ltsc2022'
 ].each { imageType ->
-    parallelStages["Building ${imageType}"] = {
+    parallelStages[imageType] = {
         withEnv(["IMAGE_TYPE=${imageType}"]) {
             int retryCounter = 0
             retry(count: 2, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
@@ -54,6 +54,7 @@ parallelStages = [failFast: false]
                 retryCounter++
                 node(resolvedAgentLabel) {
                     timeout(time: 60, unit: 'MINUTES') {
+                        checkout scm
                         if (imageType == "linux") {
                             stage('Prepare Docker') {
                                 sh 'make docker-init'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,7 +46,7 @@ parallelStages = [failFast: false]
     'windowsservercore-ltsc2022'
 ].each { imageType ->
     parallelStages[imageType] = {
-        withEnv(["IMAGE_TYPE=${imageType}"]) {
+        withEnv(["IMAGE_TYPE=${imageType}", "REGISTRY_ORG=${infra.isTrusted() ? 'jenkins' : 'jenkins4eval'}"]) {
             int retryCounter = 0
             retry(count: 2, conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()]) {
                 // Use local variable to manage concurrency and increment BEFORE spinning up any agent
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,21 @@
+final String cronExpr = env.BRANCH_IS_PRIMARY ? '@daily' : ''
+
+properties([
+    buildDiscarder(logRotator(numToKeepStr: '10')),
+    disableConcurrentBuilds(abortPrevious: true),
+    pipelineTriggers([cron(cronExpr)]),
+])
+
 def agentSelector(String imageType) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,8 +1,21 @@
...
     // Linux agent
     if (imageType == 'linux') {
-        // Need Docker and a LOT of memory for faster builds (due to multi archs) or fallback to linux (trusted.ci)
-        return 'docker-highmem || linux'
+        // This function is defined in the jenkins-infra/pipeline-library
+        if (infra.isTrusted()) {
+            return 'linux'
+        } else {
+            // Need Docker and a LOT of memory for faster builds (due to multi archs) or fallback to linux (trusted.ci)
+            return 'docker-highmem'
```
smerle33, line 17: you don't make sure you are calling a intel/AMD64 ? it can be an ARM64

dduportal: The we broke the contract as ATH is using this label (and is not expecting any arm64 agent). But the arm64 seems to come from the linux label

timja (commented):
This seems a bit over the top, why re-write it into scripted, it was a lot easier to read before?

PR description says its to do with retries but declarative handles that fine:

```
options {
  retry(conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()], count: 2)
}
```

dduportal:
> This seems a bit over the top, why re-write it into scripted, it was a lot easier to read before?
> 
> PR description says its to do with retries but declarative handles that fine:
> 
> ```
> options {
>   retry(conditions: [kubernetesAgent(handleNonKubernetes: true), nonresumable()], count: 2)
> }
> ```

Declarative syntax does parse the agent label once for all so you cannot use a dynamic label which will change over the course of the pipeline, hence the need for a scripted syntax. Additionally, you cannot access any retry counter to determine if the "non spot" label shall be used or not.

Also, most of the current pipeline is already scripted syntax, so declarative does not have a lot of value.

dduportal:
@timja @lemeurherveCB I've updated the PR body with more details about the choices and reasoning

Jenkinsfile
```diff
@@ -12,93 +25,82 @@ def agentSelector(String imageType) {
...
+    'linux',
+    'nanoserver-1809',
+    'nanoserver-ltsc2019',
+    'nanoserver-ltsc2022',
+    'windowsservercore-1809',
+    'windowsservercore-ltsc2019',
+    'windowsservercore-ltsc2022'
+].each { imageType ->
+    parallelStages[imageType] = {
+        withEnv(["IMAGE_TYPE=${imageType}"]) {
```
dduportal, line 49:
```suggestion
        withEnv(["IMAGE_TYPE=${imageType}", "REGISTRY_ORG=${infra.isTrusted() ? 'jenkins' : 'jenkins4eval'}"]) {
```

timja (approved):
Thanks!

dduportal:
I'm merging and proceeding to publish a new release (with the UBI9 change) to validate the change up to the publication. Thanks!

PR #961, opened 2025-03-26
chore(Jenkinsfile) avoid using trusted.ci agent labels on ci.jenkins.io

dduportal:
If the label `docker-highmem` label does not provide any agent on ci.jenkins.io, we don't want it to fall back on `linux` label, because it may lead to spinning up an arm64 machine which fails the build

Extracted from https://github.com/jenkinsci/docker-agent/pull/960 to provide a short term fix to failing builds, unrelated to the support of spot/retries

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,13 @@
 def agentSelector(String imageType) {
     // Linux agent
     if (imageType == 'linux') {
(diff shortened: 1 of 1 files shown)


lemeurherveCB (approved):

smerle33 (approved):
🚀

dduportal (commented):
Merging as it is only a minor infrastructure related change

PR #979, opened 2025-05-05
Bump UBI9 version to 9.5-1745848351

MarkEWaite:
## Bump UBI9 version to 9.5-1744101466

The updatecli process is failing due to some change in the UBI 9 API response.  Rather than delay the availability of remoting 3309 and the latest operating system updates, I've created the updates myself.

Includes a change that needs to be reviewed by those who understand updatecli better than I do.  @gounthar or @dduportal or @lemeurherveCB would be great to review that proposed change.  The change works, but I am not confident that it is correct.

### Testing done

Confirmed that `make build` and `make clean` are both successful.

### Submitter checklist

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -105,7 +105,7 @@ variable "DEBIAN_RELEASE" {
 }
 
 variable "UBI9_TAG" {
-  default = "9.5-1744101466"
+  default = "9.5-1745848351"
 }
 
 # Set this value to a specific Windows version to override Windows versions to build returned by windowsversions function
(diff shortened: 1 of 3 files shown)


MarkEWaite:
> Thanks, Mark.

Thanks for reviewing.  If you're in the office tomorrow, could you look at the change I made to the updatecli definition?  It provides the results that I expect, but I'm not sure that it is actually correct.

dduportal (approved):
- LGTM for the new UBI9 version (which is the core of this PR)
- LGTM for the `updatecli` script change used to detect the latest UBI9 available version: the checks (which runs against the `master` branch) says:
```
⚠ - changes detected:
	path "variable.UBI9_TAG.default" updated from "9.5-1744101466" to "9.5-1745848351" in file "docker-bake.hcl"
```

Which is what we would have wanted!

I'm merging so we get the UBI9 version: if the updatecli process needs more change we can revisit it later

PR #981, opened 2025-05-05
Update examples to use most recent JDK versions

MarkEWaite:
## Update examples to use most recent JDK versions

Update examples to Java 17.0.15

Use 21.0.7_6 instead of 21.0.3_9 in example comment.

Includes a change that needs to be reviewed by those who understand updatecli better than I do.  @gounthar or @dduportal or @lemeurherveCB would be great to review that proposed change.  The change works, but I am not confident that it is correct.

### Testing done

Copied the example output from current output samples.

### Submitter checklist

### DIFF
--- a/rhel/ubi9/Dockerfile
+++ b/rhel/ubi9/Dockerfile
@@ -9,7 +9,7 @@ ARG TARGETPLATFORM
 COPY adoptium-get-jdk-link.sh /usr/bin/local/adoptium-get-jdk-link.sh
 COPY adoptium-install-jdk.sh /usr/bin/local/adoptium-install-jdk.sh
 
-ARG JAVA_VERSION=17.0.12_7
+ARG JAVA_VERSION=17.0.15_6
 
 RUN dnf install --disableplugin=subscription-manager --setopt=install_weak_deps=0 --setopt=tsflags=nodocs --allowerasing -y \
     ca-certificates \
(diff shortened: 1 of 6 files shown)


dduportal (approved):
After updating the branch, the `updatecli` change is gone from the diff because we merged #979 .

LGTM for the documentation change!


Ping (not important, not urgent) @gounthar, @jayfranco999 @lemeurherveCB do you think you could use `updatecli` to automate this kind of documentation change (at least the JDK version updates)?

gounthar:
I'm on it. 👍

rhel/ubi9/Dockerfile
```diff
@@ -9,7 +9,7 @@ ARG TARGETPLATFORM
 COPY adoptium-get-jdk-link.sh /usr/bin/local/adoptium-get-jdk-link.sh
 COPY adoptium-install-jdk.sh /usr/bin/local/adoptium-install-jdk.sh
 
-ARG JAVA_VERSION=17.0.12_7
+ARG JAVA_VERSION=17.0.15_6
```
lemeurherveCB, line 12: Is it expected that images have been updated in this PR while it's not doc related?

dduportal: It is a "doc" related change since it is only there as a non functional string since it's overridden by https://github.com/jenkinsci/docker-agent/blob/028dd230b41bb3ee047e9b822bc737d79a911f92/docker-bake.hcl#L316.

But it should be updated as part of the java bumps by https://github.com/jenkinsci/docker-agent/blob/master/updatecli/updatecli.d/jdk17.yaml (missing targets to UBI9) OR we remove the default value of the `Dockerfile`s

Issue #1005, opened 2025-06-25
Git 2.50.0 on Windows(version `3309.v27b_9314fd1a_4-5`): `fetch` command fails with error code 130


dduportal:
### Jenkins and plugins versions report

The packaging step of the Jenkins 2.516 release failed on Windows with the error:

```
Command "git fetch --tags --force --progress -- git@github.com:jenkinsci/packaging.git +refs/heads/*:refs/remotes/origin/*" returned status code 130
```

It used the version [`3309.v27b_9314fd1a_4-5`](https://github.com/jenkinsci/docker-agent/releases/tag/3309.v27b_9314fd1a_4-5) of the nanoserver agent which features Git 2.50.0. The failure seems consistent (at least on 3 consecutive tentatives).

With a rollback to the previous version [`3309.v27b_9314fd1a_4-4`](https://github.com/jenkinsci/docker-agent/releases/tag/3309.v27b_9314fd1a_4-4) it works with no issues.

The failure happens during a pipeline `git `step which tries to clone the repository jenkinsci/packaging with SSH:
https://github.com/jenkins-infra/release/blob/001b3c2f75d35047dee7a8f599bef9e657eac205/Jenkinsfile.d/core/package#L266

Could be the same issue as mentionned by @jonesbusy in https://github.com/jenkinsci/docker-agent/pull/1001#issuecomment-2983362123.

### What Operating System are you using (both controller, and any agents involved in the problem)?

Windows Nanoserver

### Reproduction steps

```
pipeline {
  agent {
    kubernetes {
      yamlFile 'PodTemplates.d/package-windows.yaml'
    }
  }

  stages {
    stage('Test Git') {
      steps {
        container('dotnet') {
          checkout scm
          dir('test1') {
            // HTTPS (anonymous) clone
            git branch: 'master', url: 'https://github.com/jenkinsci/packaging.git'
          }
          dir('test2') {
            // SSH clone
            git branch: 'master', credentialsId: 'release-key', url: 'git@github.com:jenkinsci/packaging.git'
          }
        }
      }
    }
  }
}
```

```yaml
apiVersion: "v1"
kind: "Pod"
metadata:
  labels:
    jenkins: "slave"
    job: "package"
    #Following label is required by the NetworkPolicy managed by stable/jenkins helm chart and configured from jenkins-infra/charts
    jenkins/default-release-jenkins-agent: true
spec:
  containers:
[...]

PR #1006, opened 2025-06-25
Revert Git version on Windows from `2.50.0.windows.1` to `2.49.0.windows.1`

dduportal:
Reverts jenkinsci/docker-agent#1001

See https://github.com/jenkinsci/docker-agent/issues/1005 and https://github.com/jenkinsci/docker-agent/pull/1001#issuecomment-2983362123

### DIFF
--- a/windows/nanoserver/Dockerfile
+++ b/windows/nanoserver/Dockerfile
@@ -56,7 +56,7 @@ COPY --from=pwsh-source $PSHOME $PSHOME
 SHELL ["pwsh.exe", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 USER ContainerAdministrator
 
-ARG GIT_VERSION=2.50.0
+ARG GIT_VERSION=2.49.0
 ARG GIT_PATCH_VERSION=1
 RUN [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12 ; `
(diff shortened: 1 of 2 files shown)


lemeurherveCB:
The workarounds seem fine to me: people blocked by the issue and that can't switch to HTTPS will have to roll back anyway, equivalent to deploy a new version including 2.49.

So for everyone else (able to switch to HTTPS), having those workarounds in the release notes avoid them to deploy a new version including 2.49

MarkEWaite:
> Since we're going to describe the "known" issue in the release notes, it means: should we rollback, or wait for a fix on upstream git?
> 
> cc @gounthar (as platform SIG leader so the subject is tracked) @MarkEWaite @jonesbusy @lemeurherveCB

The fix seems to be progressing through Git for Windows rapidly.  That is consistent with how the Git for Windows maintainer has worked in the past.  If the fix is available from Git for Windows by start of day Friday, I think we could skip the revert and apply 2.50.1 on Friday (or sooner).  If we reach Friday and don't have a fix from Git for Windows, I think we need to rollback.  There are certainly use cases where users won't want to change the authentication method for their repository and there are even cases where they can't change the authentication method and have it continue to work.  For example, if submodules are defined by ssh URL in the repository, then switching to https for clone will cause the submodule clone to fail.

MarkEWaite (approved):
I approve this change, though I think it is OK to wait a few days in hopes that Git for Windows will be released soon with a fix for the issue.

PR #1046, opened 2025-09-17
Accept `$REMOTING_OPTS` containing spaces

jglick:
Amending #809. Currently it is not possible to set for example

```bash
export REMOTING_OPTS='-noReconnectAfter 1d -webSocketHeader "Cookie=x=1; y=2"'
```

(note that the space in the `Cookie` value is mandatory) since Bash will treat this as arguments

* `-noReconnectAfter`
* `1d`
* `-webSocketHeader`
* `"Cookie=x=1;`
* `y=2"`

whereas we wanted

* `-noReconnectAfter`
* `1d`
* `-webSocketHeader`
* `Cookie=x=1; y=2`

Did not test Windows behavior here, only Linux.

### DIFF
--- a/.github/workflows/update-dockerhub-description.yaml
+++ b/.github/workflows/update-dockerhub-description.yaml
@@ -10,7 +10,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - name: Update Docker Hub description for agent
-        uses: peter-evans/dockerhub-description@v4
+        uses: peter-evans/dockerhub-description@v5
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
@@ -19,7 +19,7 @@ jobs:
           repository: jenkins/agent
           readme-filepath: ./README_agent.md
       - name: Update Docker Hub description for inbound-agent
-        uses: peter-evans/dockerhub-description@v4
+        uses: peter-evans/dockerhub-description@v5
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -14,7 +14,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Install Updatecli in the runner
-        uses: updatecli/updatecli-action@v2.92.0
+        uses: updatecli/updatecli-action@v2.96.0
 
       - name: Run Updatecli in Dry Run mode
         run: updatecli diff --config ./updatecli/updatecli.d --values ./updatecli/values.github-action.yaml
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,6 +82,7 @@ def parallelStages = [failFast: false]
                                             if (isUnix()) {
                                                 sh 'make publish'
                                             } else {
+                                                powershell '& ./build.ps1 build'
                                                 powershell '& ./build.ps1 publish'
                                             }
                                         }
@@ -91,14 +92,20 @@ def parallelStages = [failFast: false]
                                 }
                             }
                         } else {
-                            stage('Build and Test') {
-                                // ci.jenkins.io builds (e.g. no publication)
+                            // ci.jenkins.io builds (e.g. no publication)
+                            stage('Build') {
                                 if (isUnix()) {
                                     sh './build.sh'
+                                } else {
+                                    powershell '& ./build.ps1 build'
+                                    archiveArtifacts artifacts: 'build-windows_*.yaml', allowEmptyArchive: true
+                                }
+                            }
+                            stage('Test') {
+                                if (isUnix()) {
                                     sh './build.sh test'
                                 } else {
                                     powershell '& ./build.ps1 test'
-                                    archiveArtifacts artifacts: 'build-windows_*.yaml', allowEmptyArchive: true
                                 }
                                 junit(allowEmptyResults: true, keepLongStdio: true, testResults: 'target/**/junit-results*.xml')
                             }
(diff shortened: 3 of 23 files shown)


lemeurherve:
@jglick not sure I'm properly testing it, but when trying your change (with of without env vars passed to `docker run`) I got the following error:
```
$ docker buildx bake --file docker-bake.hcl inbound-agent_debian_jdk17
$ docker run -it -e REMOTING_OPTS='-noReconnectAfter 1d -webSocketHeader "Cookie=x=1; y=2"' docker.io/jenkins/inbound-agent:jdk17
/usr/local/bin/jenkins-agent: 1: eval: Syntax error: "(" unexpected
```

Passing the env var like above (which may be wrong?) to docker seems to already do the job:
From the `master` branch with `set -x` and a debug statement, I got this:
```diff
--- a/jenkins-agent
+++ b/jenkins-agent
@@ -42,6 +42,8 @@
 #                              the agent skips connecting to an HTTP(S) port for connection info.
 # * JENKINS_PROTOCOLS:         Specify the remoting protocols to attempt when instanceIdentity is provided.
 
+ set -x
+
 if [ $# -eq 1 ] && [ "${1#-}" = "$1" ] ; then
 
        # if `docker run` only has one arguments and it is not an option as `-help`, we assume user is running alternate command like `bash` to inspect the image
@@ -136,6 +138,8 @@ else
        #TODO: Handle the case when the command-line and Environment variable contain different values.
        #It is fine it blows up for now since it should lead to an error anyway.
 
+       echo "DEBUG: REMOTING_OPTS=[$REMOTING_OPTS]"
+
         exec $JAVA_BIN $JAVA_OPTIONS -jar $AGENT_FILE $SECRET $AGENT_NAME $TUNNEL $URL $WORKDIR $WEB_SOCKET $DIRECT $PROTOCOLS $INSTANCE_IDENTITY $REMOTING_OPTS "$@"
 
 fi
```
```
$ docker buildx bake --file docker-bake.hcl inbound-agent_debian_jdk17
$ docker run -it -e REMOTING_OPTS='-noReconnectAfter 1d -webSocketHeader "Cookie=x=1; y=2"' docker.io/jenkins/inbound-agent:jdk17
+ [ 0 -eq 1 ]
+ [ ! -z  ]
+ [ ! -z  ]
+ [ -n  ]
+ [ -n  ]
+ [  = true ]
+ [ -n  ]
+ [ -n  ]
+ [ -n  ]
+ [  ]
+ JAVA_BIN=java
+ [ /opt/java/openjdk ]
+ JAVA_BIN=/opt/java/openjdk/bin/java
+ [  ]
+ [  ]
+ [  ]
+ AGENT_FILE=/usr/share/jenkins/agent.jar
+ [ -n  ]
+ [ -n  ]
+ echo DEBUG: REMOTING_OPTS=[-noReconnectAfter 1d -webSocketHeader "Cookie=x=1; y=2"]
DEBUG: REMOTING_OPTS=[-noReconnectAfter 1d -webSocketHeader "Cookie=x=1; y=2"]
+ exec /opt/java/openjdk/bin/java -jar /usr/share/jenkins/agent.jar -noReconnectAfter 1d -webSocketHeader "Cookie=x=1; y=2"
option "-webSocketHeader" requires the option(s) [-webSocket]
java -jar agent.jar [options...]
<snip>
```
This looks good to me, then I may be missing other elements and/or context.

Do you have steps to reproduce your errors by any chance?

jglick:
If I could build images, probably https://github.com/jenkinsci/docker-agent/blob/329ea1c948b7089b4282ed9c5f849fef1f0b52c8/tests/tests_inbound-agent.bats#L26-L43 could be duplicated to run `-webSocket -webSocketHeader "Cookie=x=1; y=2"` and assert that this line is received by netcat.

MarkEWaite:
> Is there a reason not to use just use the Java packages provided by the various distros?

We've preferred to have a single distribution that we use across all supported platforms and all supported operating systems.  Eclipse Temurin provides that.  They also use Jenkins in the creation of their releases and we like that as well.

Eclipse Temurin builds pass the [Java Technology Compatibility Kit](https://en.wikipedia.org/wiki/Technology_Compatibility_Kit) certification and their own [Aquavit Framework](https://adoptium.net/aqavit).

Eclipse Temurin provides a Windows distribution that avoids us having to choose which other Java vendor we would use for our Windows container images.

Eclipse Temurin 21+ provides [reproducible builds](https://adoptium.net/news/2024/08/adoptium-reproducible-verification-builds).

gounthar:
I've noticed that not all Linux distributions currently provide Temurin binaries, which we actively support. Additionally, there seems to be a delay in the availability of OpenJDK builds from several distributions.

tests/tests_inbound-agent.bats
```diff
@@ -42,6 +42,17 @@ SUT_IMAGE="$(get_sut_image)"
   cleanup "${sut_cid}"
 }
 
+@test "[${SUT_IMAGE}] image parses \$REMOTING_OPTS correctly (slow test)" {
+  local netcat_cid sut_cid
+  netcat_cid="$(docker run -d -it netcat-helper:latest /bin/sh -c "while :; do (echo 'HTTP/1.1 200 OK'; echo) | nc -l 5000; done")"
+  sut_cid="$(docker run -d --link "${netcat_cid}" -e REMOTING_OPTS="-url http://${netcat_cid}:5000 -name xxx -secret xxx -webSocket -webSocketHeader \"Cookie=x=1; y=2\"" "${SUT_IMAGE}")"
+  sleep 5
+  run docker logs "${netcat_cid}"
+  echo "${output}" | grep 'Cookie:x=1; y=2'
```
jglick, line 51: Note that when this fails, the actual output from either container is not visible. But I can confirm that in trunk the SUT container prints

```
Two arguments required, but got [y=2"]
java -jar agent.jar [options...]
 -agentLog FILE                         : Local agent error log destination
…
```

dduportal (approved):
Looks good to me! Thanks!

PR #1066, opened 2025-10-03
chore(linux): no dot in jlink version switch

lemeurherve:
This PR simplify a bit the jlink version switch so we don't have to set a dot or not depending on the version status (GA with only a number, GA with a number and a dot, or beta with only a number).

### Testing done

`make test`

### Submitter checklist

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -41,11 +41,10 @@ RUN apk add --no-cache \
 
 ENV PATH="/opt/jdk-${JAVA_VERSION}/bin:${PATH}"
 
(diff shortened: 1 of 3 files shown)


gounthar (approved):

timja (approved):
Just removing the `.` would probably have been simpler and consistent across all versions? but this works

lemeurherve:
> Just removing the `.` would probably have been simpler and consistent across all versions? but this works

Removing the dot from `jlink --version` output?

I went the `cut` way to avoid conflicts with potential other parts of the version matching with 17, 21 or 25.

PR #1068, opened 2025-10-03
chore(windows): run each JDK tests in parallel

lemeurherve:
WIP, not for review yet.

This PR allows to run each JDK tests in parallel instead of one JDK after one JDK, reducing the time to run all Windows images tests.

<details><summary>Before</summary>

image

https://ci.jenkins.io/job/Packaging/job/docker-agent/job/master/1000/

</details>

<details><summary>After (WIP)</summary>

image
https://ci.jenkins.io/job/Packaging/job/docker-agent/job/master/1000/

</details>

Related:
- https://github.com/jenkinsci/docker/issues/1149

### Testing done

CI

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

PR #1071, opened 2025-10-05
chore(pipeline): split `Build and Test` stage

lemeurherve:
This PR splits the existing stage into "Build" and "Test".

Having separate stages allows a better understanding of the potential gains on each of those part when improving the building process.

Notes:
- Generated docker compose files are archived even if the tests fail
- As the `publish` target includes building the images, no need to call it first when only that one is running on trusted.ci.jenkins.io
- As there is no "test" phase for multi-arch build, no need to change it

Extracted from:
- #1068

### Testing done

- CI

### Submitter checklist

PS: easier to review without whitespaces, see https://github.com/jenkinsci/docker-agent/pull/1071/files?diff=unified&w=1

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -91,14 +91,20 @@ def parallelStages = [failFast: false]
                                 }
                             }
                         } else {
-                            stage('Build and Test') {
-                                // ci.jenkins.io builds (e.g. no publication)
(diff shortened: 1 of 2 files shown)


MarkEWaite (approved):
Thanks!

PR #1159, opened 2026-02-08
chore(pipeline): facilitate replay of specific bake targets

lemeurherve:
This change allows to more easily replay publication of specific bake targets in case of missing tags for example.
It also ensure that the build before publication on trusted.ci.jenkins.io is a full multiarch one so the publication step is only taking care of pushing built layers to Docker Hub.

For a replay to publish only 2 inbound agent bake targets if their tags were not published due to hypothetic Docker Hub issues:
```diff
--- old/Jenkinsfile
+++ new/Jenkinsfile
@@ -42,11 +42,8 @@
 // Specify parallel stages
 // Linux: bake group(s) or target(s)
 // Windows: flavor and version to build
 def parallelStages = [failFast: false]
 [
-    'linux',
-    'nanoserver-ltsc2019',
-    'nanoserver-ltsc2022',
-    'windowsservercore-ltsc2019',
-    'windowsservercore-ltsc2022'
+    'inbound-agent_debian_jdk17',
+    'inbound-agent_debian_jdk21',
 ].each { imageType ->
 ```

Amends:
- #1156

Ref:
- Thread on `#jenkinsci/release`: https://matrix.to/#/!JlkqzpdEnsUUuVtjgE:matrix.org/$wV5FXxmOAqFy0I5O14RH2iVADaeTHuEbrRQ_TAYP_RQ?via=matrix.org&via=gitter.im

### Testing done

- ci.jenkins.io: CI
- trusted.ci.jenkins.io: replay https://trusted.ci.jenkins.io/job/Containers/job/Agents/job/docker-agents/view/tags/job/3355.v388858a_47b_33-12/5 to publish missing agent debian jdk17 & jdk21
  <details><summary>diff</summary>

  ```diff
  --- old/Jenkinsfile
  +++ new/Jenkinsfile
  @@ -42,11 +42,8 @@
   // Specify parallel stages
   // Linux: bake group(s) or target(s)
   // Windows: flavor and version to build
   def parallelStages = [failFast: false]
   [
  -    'linux',
  -    'nanoserver-ltsc2019',
  -    'nanoserver-ltsc2022',
  -    'windowsservercore-ltsc2019',
  -    'windowsservercore-ltsc2022'
  +    'agent_debian_jdk17',
  +    'agent_debian_jdk21',
   ].each { imageType ->
   ```

  </details>

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
[...]

PR #1161, opened 2026-02-15
fix(pipeline): pass publication env vars to all stages

lemeurherve:
This change amends #1159 restoring missing Windows tags with the remoting version and build number by passing `ON_TAG`, `REMOTING_VERSION` and `BUILD_NUMBER` to all stages, with default values when not in a tag build.

Ref:
- https://github.com/jenkinsci/docker-agents/issues/1160#issuecomment-3902896199

### Testing done

Replay on trusted.ci.jenkins.io setting those values as soon as the "docker init" stage, where make.ps1 is generating with `docker bake` the docker compose file used later to build and publish the Windows images: https://trusted.ci.jenkins.io/job/Containers/job/Agents/job/docker-agents/view/tags/job/3355.v388858a_47b_33-13/2

### Submitter checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,6 +39,20 @@ def agentSelector(String imageType, retryCounter) {
     ])
 }
 
+// Defaul values
+def tagWithOneDashExist = false
+def remotingVersion = '3355.v388858a_47b_33'
+def buildNumber = env.BUILD_NUMBER
+// Values on tag containing the remoting version and build number
+if (env.TAG_NAME) {
+    def tagItems = env.TAG_NAME.split('-')
+    if (tagItems.length == 2) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -39,6 +39,20 @@ def agentSelector(String imageType, retryCounter) {
     ])
 }
 
+// Defaul values
+def tagWithOneDashExist = false
+def remotingVersion = '3355.v388858a_47b_33'
+def buildNumber = env.BUILD_NUMBER
+// Values on tag containing the remoting version and build number
+if (env.TAG_NAME) {
```
timja: I'm surprised this is set, I'd expect it to need a checkout step to be run first

PR #1166, opened 2026-02-19
fix(pipeline): use `infra.withDockerCredentials` in all stages

lemeurherve:
This change ensures we're logged in Docker Hub in all stages and all controllers, not only the publication one on trusted.ci.jenkins.io

Best reviewed without whitespace: https://github.com/jenkinsci/docker-agents/pull/1166/changes?diff=unified&w=1

Ref:
- https://github.com/jenkinsci/docker-agents/issues/1163#issuecomment-3926946897

### Testing done

CI

I'll also test a publication on trusted.ci without #1162 to see if we still encounter 429 rate limit errors.

### Submitter checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -83,65 +83,65 @@ def parallelStages = [failFast: false]
                 node(resolvedAgentLabel) {
                     timeout(time: 60, unit: 'MINUTES') {
                         checkout scm
-                        stage("Prepare Docker on ${resolvedAgentLabel}") {
-                            if (isUnix()) {
-                                sh 'make docker-init'
-                            } else {
-                                powershell './make.ps1 docker-init'
-                                archiveArtifacts artifacts: 'build-windows_*.yaml', allowEmptyArchive: true
-                            }
-                        }
-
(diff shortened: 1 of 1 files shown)


dduportal (commented):
We can merge this if it helps on short term.


BUT, there is something weird: we have a mirror registry set up in the `daempon.json` of each Docker Engine in our VM agents (unsecure homemade registry in AWS AKS for ci.io and ACR for trusted.ci/release.ci/infra.ci/cert.ci).

No authentication for pull should be setup as the Docker Engine is expected to only discuss with the mirror registry. The mirror registry is set up with a DockerHub credential to access the DocerkHub.

However, if the Docker Engine can't reach the mirror registry, then it falls back to DockerHub (which would indicates something is wrong and must be fixed). 

We chose (Stephane, Mark and I, I remember Tim also agreeing but it was many months ago) to willingly remove the pull credential here to ensure we "see" something because monitoring the Docker Engine fallback from mirror to DocerkHub ain't easy (except collecting Docker daemon logs and adding alerts based on pattern, which is a LOT of logs given our build rate)

PR #1263, opened 2026-07-22
chore(windows): switch Adoptium installation to ZIP archive instead of MSI

dduportal:
Ref. https://github.com/jenkinsci/docker-agents/issues/1265

This PR changes the installation method of Adoptium inside our Docker containers.

- Only for Windows! Linux would be next but let's go step by step.
- No GPG verification (yet): we don't do it today but it could  be a great improvement (https://adoptium.net/installation/gpg-verification)
- Introduces a new JSON `docker-bake.override.json` file as it's easier to keep up to date with tools like updatecli or renovabot. Had to add it to all calls to `docker buildx bake` CLI (even Linux in advance to avoid forgetting about it).
  - I guess we could remove all `--file` flags for buildx bake and let https://docs.docker.com/build/bake/reference/#file-format do its job. Worth another PR for this.
- Removes the `JAVA_VERSION`  build argument (unneeded anymore and not passed to end user as `ENV` so no user facing change)
  - Includes test updates to handle this change

### Testing done

- Was able to get my build stuck on my Windows 11 machine (building the 2019 images) as preliminary
- Ran locally the build: passed without being stuck
- Locally running tests: java version are 👍 (but stuck on nmap)
- Checking ci.jenkins.io: same behavior

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ check_cli = type "$(1)" >/dev/null 2>&1 || { echo "Error: command '$(1)' require
 ## Check if a given image or group exists in the current manifest docker-bake.hcl
 check_image = $(MAKE) --silent list listgroup-all | grep -w '$(1)' >/dev/null 2>&1 || { echo "Error: the image or group '$(1)' does not exist in manifest for the current platform '$(OS)/$(ARCH)'. Please check the output of '$(MAKE) list' or '$(MAKE) listgroup-all'. Exiting." ; exit 1 ; }
 ## Base "docker buildx base" command to be reused everywhere
-bake_base_cli := docker buildx bake --file docker-bake.hcl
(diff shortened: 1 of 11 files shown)


dduportal:
Hello @timja @MarkEWaite Could I get a review before I force merge?

- CI checks are expected to fail due to timeout building the nmap container (but build + tests related to JDKs should pass)
- The goal is to consider https://github.com/jenkinsci/docker-agents/releases/tag/3383.vc8881d4b_0e76-2 dead and quickly release a 3383.vc8881d4b_0e76-3 (before moving forward with new remoting version).
- I'll open a subsequent PR with updatecli (already tested locally) to track the versions of JDKs for windows

MarkEWaite (approved):

PR #1266, opened 2026-07-24
tests(windows) replace nmap container by a pure powershell TCP listener running inside a windowsserver container

dduportal:
Ref. https://github.com/jenkinsci/docker-agents/issues/1265

This PR gets rid of the NMAP custom container rebuilt on each run on Windows. This container is stuck  while being built on Windows 2019 and is slowing down the processes.

I chose to keep the "use a container" pattern despite it being slow (container startup overhead) to avoid introducing too much changes at a time: the goal is to unblock the stuck CI checks right now. But we could even get rid of the container pattern for the TCP listener and use a pure powershell function with a background task, at the risk of depending on the Windows host firewall (which could annoys contributors). Tradeoff.

Important: this PR was largely vibe coded with Claude Code so might need refinement. I rely on the CI checks for now.


tests/inbound-agent.Tests.ps1
```diff
@@ -83,36 +81,29 @@ Describe "[$global:IMAGE_NAME] image has jenkins-agent.ps1 in the correct locati
...
+        $exitCode, $stdout, $stderr = Run-Program 'docker' "run --detach --tty --network=jnlp-network --name $global:CONTAINERNAME $global:IMAGE_NAME -Url http://${ncat_ip}:5000 $secret $name"
         $exitCode | Should -Be 0
         Is-ContainerRunning $global:CONTAINERNAME | Should -BeTrue
 
-        $exitCode, $stdout, $stderr = Run-Program 'docker' 'wait nmap'
-        $exitCode, $stdout, $stderr = Run-Program 'docker' 'logs nmap'
+        $exitCode, $stdout, $stderr = Run-Program 'docker' 'wait ncat-helper'
+        $exitCode, $stdout, $stderr = Run-Program 'docker' 'logs ncat-helper'
         $exitCode | Should -Be 0
         $stdout | Should -Match "GET /tcpSlaveAgentListener/ HTTP/1.1`r"
```
dduportal:
```suggestion
        $stdout | Should -Match "GET /tcpSlaveAgentListener/ HTTP/1.1"
```

dduportal:
Tests passed, yeaaaah

PR #1267, opened 2026-07-26
chore(updatecli) track Adoptium Windows Installer URLs

dduportal:
Follow up of #1263 

This PR tracks the Windows Adoptium installed URLs introduced in #1263 to bump with new versions when available


dduportal:
Self-merging: GHA build output looks good!

PR #1272, opened 2026-07-29
chore: switch to Adoptium archives installers instead of custom scripts

dduportal:
This PR changes all images to install Adoptium from their archive installers.

- It's mainly targeted to Linux images as Windows images were treated in #1263
- Updatecli tracking will be broken but fixed in a subsequent PR already opened to verify all assumptions: https://github.com/jenkinsci/docker-agents/pull/1281
- **However** after discussions with @lemeurherve , following his suggestion, we changed the Windows images to follow the same pattern: we have a single text file with all JDK/OS/CPU installer URLs as a reference and this file is copied and read inside `Dockerfile`. Implies a few things:
  - Mainly motivated to avoid too much changes in the buildx bake tooling: supporting many CPU architectures (Linux only) meant we would need to add the `platform` as a 3rd matrix dimension in the bake targets which requires a lot of changes not really required.
  - Removal of the `docker-bake.override.json` file introduced in #1263 and associated `docker buildx bake --file *` flags.
  - TXT file instead of YML/JSON requires no additional tools such as `jq` or `yq`. Easier and less fragile at the expense of a "raw" text file.
- Cleanup of tooling related to the "JAVA_VERSION" which is not known anymore as build `ARG` (less parsing logic).
- The old `adoptium*sh` script are removed and not tracked anymore \o/

### Submitter checklist


alpine/Dockerfile
```diff
@@ -49,7 +54,7 @@ RUN java_major_version="$(jlink --version 2>&1 | cut -c1-2)"; \
     # because jlink fails with the error "jmods: Value too large for defined data type" error.
     # For JDK25, we need to have a proper test and validation process first cf https://github.com/jenkinsci/docker-agents/issues/1124)
     # Noting there isn't any JMODs shipped anymore since https://adoptium.net/en-GB/news/2025/08/eclipse-temurin-jdk24-JEP493-enabled
-    cp -r "/opt/jdk-${JAVA_VERSION}" /javaruntime; \
+    mv "${temp_Java_Home}" /javaruntime; \
```
dduportal: - Faster to move than copying on a non-overlay filesystem. Not sure if it really changes anything in this overlay context (as the underlying backing filesystem will most probably just link the bits when performing a copy).
- But ensures the original JDK is not there anymore to avoid confusion

alpine/Dockerfile
```diff
@@ -28,18 +28,23 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 # Ref. https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
 ARG TARGETPLATFORM
 
-COPY adoptium-get-jdk-link.sh /usr/bin/local/adoptium-get-jdk-link.sh
-COPY adoptium-install-jdk.sh /usr/bin/local/adoptium-install-jdk.sh
-
-ARG JAVA_VERSION
+ARG JAVA_RELEASE
+COPY ./adoptium/jdk_installers.txt /tmp/jdk_installers.txt
+ARG temp_Java_Home=/opt/jdk
```
lemeurherve:
```suggestion
ARG TEMP_JAVA_HOME=/opt/jdk
```
More consistent case.

If that value is never set to anything else, I'd suggest removing that ARG altogether and just use a local variable in the RUN block.

alpine/Dockerfile
```diff
@@ -28,18 +28,23 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
...
     ca-certificates \
-    jq \
     curl \
-  && /usr/bin/local/adoptium-install-jdk.sh alpine
-
-ENV PATH="/opt/jdk-${JAVA_VERSION}/bin:${PATH}"
+    tar \
+  && java_zip_url="$(grep "jdk${JAVA_RELEASE}/${TARGETPLATFORM}/musl" /tmp/jdk_installers.txt | cut -d' ' -f2)" \
+  && mkdir -p "${temp_Java_Home}" \
+  && jdkArchive="/opt/jdk.zip" \
```
lemeurherve: camelCase inconsistent with the rest of the RUN script, suggesting to use snake_case

debian/Dockerfile
```diff
@@ -28,19 +28,24 @@ SHELL ["/bin/bash", "-e", "-u", "-o", "pipefail", "-c"]
 # Ref. https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
 ARG TARGETPLATFORM
 
-COPY adoptium-get-jdk-link.sh /usr/bin/local/adoptium-get-jdk-link.sh
-COPY adoptium-install-jdk.sh /usr/bin/local/adoptium-install-jdk.sh
-
-ARG JAVA_VERSION
+ARG JAVA_RELEASE
```
lemeurherve, line 31: To be discussed: suggesting renaming this to `JDK` here, and something similar in docker-bake, to be consistent with the existing `jdk`<something> variables and functions already there.

docker-bake.hcl
```diff
@@ -96,7 +81,7 @@ target "alpine" {
   args = {
     ALPINE_TAG   = ALPINE_FULL_TAG
     VERSION      = REMOTING_VERSION
-    JAVA_VERSION = "${javaversion(jdk)}"
+    JAVA_RELEASE = jdk
```
lemeurherve, line 84: See above

lemeurherve (dismissed):
Nice 🙌 

Some naming suggestions.

alpine/Dockerfile
lemeurherveCB: Note: I don't mind opening a follow-up PR dedicated to this renaming, including the Windows parts too at the same time.

dduportal: - Ack for the case (suggestion): since the variable is available as a shell environment variable it makes sense to use uppercase
- The value is used to define the `ENV PATH` below. So we cannot set it up as a local RUN variable otherwise it won't be available (and we will have to duplicate the value which defeat the whole purpose of using a variable here)

alpine/Dockerfile
dduportal:
```suggestion
  && jdk_archive="/opt/jdk.zip" \
```

alpine/Dockerfile
```diff
@@ -28,18 +28,23 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
...
+  && curl --silent --show-error --location --output "${jdkArchive}" "${java_zip_url}" \
+  # TODO: Verify download with GPG
+  && tar xzf "${jdkArchive}" --strip-components=1 -C "${temp_Java_Home}" \
```
dduportal:
```suggestion
  && curl --silent --show-error --location --output "${jdk_archive}" "${java_zip_url}" \
  # TODO: Verify download with GPG
  && tar xzf "${jdk_archive}" --strip-components=1 -C "${TEMP_JAVA_HOME}" \
```

Ref. https://github.com/jenkinsci/docker-agents/pull/1272/changes#r3722275280 and https://github.com/jenkinsci/docker-agents/pull/1272/changes#r3722266938

alpine/Dockerfile
```diff
@@ -28,18 +28,23 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
...
 ARG TARGETPLATFORM
 
-COPY adoptium-get-jdk-link.sh /usr/bin/local/adoptium-get-jdk-link.sh
-COPY adoptium-install-jdk.sh /usr/bin/local/adoptium-install-jdk.sh
-
-ARG JAVA_VERSION
+ARG JAVA_RELEASE
+COPY ./adoptium/jdk_installers.txt /tmp/jdk_installers.txt
+ARG temp_Java_Home=/opt/jdk
+ENV PATH="${temp_Java_Home}/bin:${PATH}"
```
dduportal:
```suggestion
ENV PATH="${TEMP_JAVA_HOME}/bin:${PATH}"
```

alpine/Dockerfile
```diff
@@ -28,18 +28,23 @@ SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
...
 RUN apk add --no-cache \
     ca-certificates \
-    jq \
     curl \
-  && /usr/bin/local/adoptium-install-jdk.sh alpine
-
-ENV PATH="/opt/jdk-${JAVA_VERSION}/bin:${PATH}"
+    tar \
+  && java_zip_url="$(grep "jdk${JAVA_RELEASE}/${TARGETPLATFORM}/musl" /tmp/jdk_installers.txt | cut -d' ' -f2)" \
+  && mkdir -p "${temp_Java_Home}" \
```
dduportal:
```suggestion
  && mkdir -p "${TEMP_JAVA_HOME}" \
```

alpine/Dockerfile
dduportal:
```suggestion
    mv "${TEMP_JAVA_HOME}" /javaruntime; \
```

debian/Dockerfile
dduportal: I'm not sure of the value of using consistent naming since the bakefile and the `Dockerfile` have different (abstract) contexts. The choice of `JAVA_RELEASE` comes from the official Adoptium API attributes naming: the major java version is referenced as the "Java Release".

If we want consistency (I don't mind to do it, does not change anything for the end users and easy to do in this PR), then I would prefer getting rid of `jdk` which is really confusing: is it the JDK distribution? version? type?

lemeurherve: Make sense. Maybe doing that change in a follow-up PR would be better though, as it would touch parts unrelated to the goal of this PR.

dduportal: > Maybe doing that change in a follow-up PR would be better though, as it would touch parts unrelated to that PR goal.

🤔 good point. I'll propose something after this one (and the associated updatecli) is merged

rhel/ubi9/Dockerfile
```diff
@@ -7,20 +7,24 @@ SHELL ["/bin/bash", "-e", "-u", "-o", "pipefail", "-c"]
...
     ca-certificates \
     curl \
-    jq \
-    && dnf clean --disableplugin=subscription-manager all \
-    && /usr/bin/local/adoptium-install-jdk.sh
-
-ENV PATH="/opt/jdk-${JAVA_VERSION}/bin:${PATH}"
+    tar \
+  && dnf clean --disableplugin=subscription-manager all \
+  && java_zip_url="$(grep "jdk${JAVA_RELEASE}/${TARGETPLATFORM}/glibc"  /tmp/jdk_installers.txt | cut -d' ' -f2)" \
```
lemeurherve:
```suggestion
  && java_zip_url="$(grep "jdk${JAVA_RELEASE}/${TARGETPLATFORM}/glibc" /tmp/jdk_installers.txt | cut -d' ' -f2)" \
```
nit

windows/nanoserver/Dockerfile
lemeurherve:
```suggestion
    $jdkTemp = 'C:\jdktemp' ; `
    $zip = Join-Path $jdkTemp 'jdk.zip' ; `
    $urls = Select-String -Path ".\jdk_installers.txt" -Pattern "jdk${env:JAVA_RELEASE}/windows/amd64" ; `
    $url = $urls.Split(' ')[1] ; `
    Invoke-WebRequest $url -OutFile $zip ; `
    Expand-Archive -Path $zip -DestinationPath $jdkTemp ; `
    $expandedDir = Get-ChildItem -Path $jdkTemp -Directory | Where-Object { $_.Name -like 'jdk*' } ; `
   Move-Item -Path $expandedDir.FullName -Destination $env:JAVA_HOME -Force ; `
   Remove-Item -Path $jdkTemp -Recurse | Out-Null
```
No need for env var there.
Renamed a bit like what's already there since "`pwsh` installed from zip" merge.

windows/nanoserver/Dockerfile
```diff
@@ -28,12 +28,15 @@ FROM mcr.microsoft.com/windows/servercore:"${WINDOWS_VERSION_TAG}" AS jdk-pwsh
 # $ProgressPreference: https://github.com/PowerShell/PowerShell/issues/2138#issuecomment-251261324
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
-ARG JAVA_HOME
-ARG JAVA_ZIP_URL="Provided by docker-bake.hcl"
+ARG JAVA_RELEASE
+ENV JAVA_HOME="C:/openjdk-${JAVA_RELEASE}"
+COPY ./adoptium/jdk_installers.txt "C:\jdk_installers.txt"
 ARG jdkTemp="C:\jdktemp"
 ARG localArchive="${jdkTemp}\jdk.zip"
```
lemeurherve:
```suggestion
```
cf previous suggestion

windows/windowsservercore/Dockerfile
```diff
@@ -28,12 +28,15 @@ FROM mcr.microsoft.com/windows/servercore:"${WINDOWS_VERSION_TAG}" AS jdk-core
 # $ProgressPreference: https://github.com/PowerShell/PowerShell/issues/2138#issuecomment-251261324
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
-ARG JAVA_HOME
-ARG JAVA_ZIP_URL="Provided by docker-bake.hcl"
+ARG JAVA_RELEASE
+ENV JAVA_HOME="C:/openjdk-${JAVA_RELEASE}"
+COPY ./adoptium/jdk_installers.txt "C:\jdk_installers.txt"
 ARG jdkTemp="C:\jdktemp"
```
lemeurherve: Same as nanoserver

windows/nanoserver/Dockerfile
dduportal: Oh, I see: `Join-Path` should do the trick! I was facing double-interpolation problems when trying to do a `$zip = "${jdkTemp}/jdk.zip"` initially, and had to use env. vars through `ARG.

lemeurherve (approved):
🚀

debian/Dockerfile
dduportal: https://github.com/jenkinsci/docker-agents/pull/1313

PR #1276, opened 2026-08-03
Update bake platform tooling

dduportal:
Extracted from https://github.com/jenkinsci/docker-agents/pull/1272, this PR is kept as draft to facilitate the review/brainstorm with @lemeurherve and other maintainers.

After discussion with Herve, it appears that an initial (and maybe final) implementation of the "use Adoptium ZIP and only maintain download URLs instead of tedious shell script" in #1272 would be way easier by copying the JSON file with the URLs (per OS/platform/JDK line) in the parent image and resolve the current CPU platform  within the build containers.


dduportal:
Not needed anymore

PR #1313, opened 2026-08-15
chore: use 'java relase' as a coherent naming with Adoptium API

dduportal:
Ref. https://github.com/jenkinsci/docker-agents/pull/1272#discussion_r3727360398

This PR only changes the naming of the "major java version" across our tools (bake file and windows tests) to ensure we get a coherent naming with the build arg. `JAVA_RELEASE` introduced in #1272 and the Adoptium API (see the `jdk*.yaml` updatecli manifest).

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -3,14 +3,14 @@ variable "agent_types_to_build" {
   default = ["agent", "inbound-agent"]
 }
 
-variable "jdks_to_build" {
+variable "java_releases_to_build" {
   default = [21, 25]
 }
-variable "default_jdk" {
+variable "default_java_release" {
   default = 21
 }
 
-variable "jdks_in_preview" {
+variable "java_releases_in_preview" {
   default = []
 }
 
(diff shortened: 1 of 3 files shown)


timja (approved):

docker-bake.hcl
```diff
@@ -217,12 +217,6 @@ function "windowsagenttypes" {
   : agent_types_to_build)
 }
 
-# Return an array of RHEL UBI 9 platforms to use depending on the jdk passed as parameter
-function "rhel_ubi9_platforms" {
-  params = [jdk]
-  result = ["linux/amd64", "linux/arm64", "linux/ppc64le"]
-}
```
lemeurherve, line 224: Not needed anymore?

dduportal: It was not needed since we dropped JDK17 (at least?): the goal of this function (like it former counterpart removed when dropping JDK17) was to return a list of supported architectures per JDK.

Now, all architectures are built on all JDKs, so the function is useless.
I chose to remove it as part of this PR since it's literally using the "JDK" former naming.

See https://github.com/jenkinsci/docker-agents/pull/1313/changes#diff-870f6fe23fc034f008f5203ee1e628d7bfa65a49095d5a4688db498328449ed2L123 (where I remove the only call to the function by the static list)

PR #1338, opened 2026-09-14
chore(debian/alpine): specify Docker Hub registry mirror on ci.jenkins.io

lemeurherve:
Refs:
- https://github.com/jenkins-infra/helpdesk/issues/5274#issuecomment-5583238525
- https://github.com/jenkinsci/docker-agents/issues/1163

### Testing done

CI

### Submitter checklist


Jenkinsfile
```diff
@@ -43,6 +43,9 @@ if (env.TAG_NAME) {
     }
 }
 
+// Docker Hub mirror for ci.jenkins.io
+def registryMirrorPrefix = 'https://k8s-hubmirro-hubmirro-477f7dfd76-9bbe4e560ee83036.elb.us-east-2.amazonaws.com:5000/v2/library/'
```
dduportal, line 47: That's not the registry you want to use ;)

We want to use the AWS-managed public ECR which mirrors all DockerHub official images: https://docs.aws.amazon.com/fr_fr/AmazonECR/latest/public/docker-pull-ecr-image.html

- See their blog post when they introduced it: https://aws.amazon.com/blogs/containers/docker-official-images-now-available-on-amazon-elastic-container-registry-public/
- Requires authentication (a bit of setup on the ec2 fleet on terraform side but already set up for the EKS node so the IAM setup should be easy) to avoid rate limit

